PR #3, opened 2019-02-13
Move all non-dynamic styling into CSS and made miscellaneous style changes

connorbo97:
- Moves React in-line stylings that are not dynamic into CSS files
- Fixes issue with panel 'crunching' when you open and close it
- When closing/opening panel, dropdown no longer shifts left and right
- Added margin to Codemirrror-lines to ensure cursor doesn't get cut off
- Made style consistent between Login.js and CreateUser.js
     -Removed CreateUser.css, as now CreateUser uses Login.css
- Refactored/Split source doc strings into functions in Output.js


src/components/CreateUser.js
```diff
@@ -133,156 +139,95 @@ class CreateUser extends React.Component {
...
+          waiting={this.state.waiting}
+          onChange={this.updateUsername}
+        />
+        <LoginInput
+          type={"password"}
+          data={this.state.password}
+          waiting={this.state.waiting}
+          onChange={this.updatePassword}
+        />
+        {this.renderErrorMessage(this.state.passwordMessage)}
```
malsf21: @connorbo97 should we explicitly do `this.renderErrorMessage(this.state.passwordMessage, false)`?

src/components/CreateUser.js
```diff
@@ -133,156 +139,95 @@ class CreateUser extends React.Component {
...
-              )}
-              {this.renderButton()}
-              <Link to="/login" className="create-form-link">
-                Already have an account? Click here to log in
-              </Link>
-            </div>
-          </div>
+      <div className="login-modal">
+        <form className="login-form" onSubmit={this.submit}>
+          {" "}
```
malsf21: Is there a reason that this is here? Is it the modal hack thing? @connorbo97

connorbo97: Yup, modal hack. It's probably a good idea to investigate why its needed/remove it. I'll make a card in trello bout it.

src/components/CreateUser.js
malsf21: Upon further consideration, this is fine with the specialized scope of the function.

PR #6, opened 2019-02-21
Added save button

aseem191:


src/components/Editor/components/Main.js
```diff
@@ -17,6 +18,8 @@ class Main extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
+      paneStyle: { transition: "none" },
```
connorbo97: Don't think this style affects anything. When I ctrl-f there's only one instance of it. There was an earlier change that removed a lot of the inline styling so it was probably a mis-merge

src/components/Editor/components/Main.js
```diff
@@ -28,12 +31,34 @@ class Main extends React.Component {
...
+  };
+
+  handleSave = event => {
+    var programsJson = {};
+
+    programsJson["HTML"] = this.props.programs.getIn(["HTML", "code"]);
+    programsJson["Processing"] = this.props.programs.getIn(["Processing", "code"]);
+    programsJson["Python"] = this.props.programs.getIn(["Python", "code"]);
+
+    fetch.updatePrograms(this.props.uid, programsJson).then(() => {
```
connorbo97: For now I think this is fine, but we should change this to check if there was an error/it failed to save. Maybe we can make the button red and make the text 'Failed to save'

src/components/Editor/components/SaveButton.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+
+/**-------Props--------
+ * runCode: function to be called when button is pressed
+ */
+const RunButton = props => (
```
connorbo97: Bc we're making a common Button later, no need to change, but in the future remember to update the component name!

src/components/Editor/components/Main.js
connorbo97: For now I think this is fine, but we should change this to check if there was an error/it failed to save. Maybe we can make the button red and make the text 'Failed to save'

src/components/Editor/components/Main.js
```diff
@@ -3,7 +3,8 @@ import SplitPane from "react-split-pane";
 import OutputContainer from "../containers/OutputContainer.js";
 import TextEditorContainer from "../containers/TextEditorContainer";
 import DropdownButton from "./DropdownButton";
-import RunButton from "./RunButton";
+import SaveButton from "./SaveButton";
+import * as fetch from "../../../lib/fetch.js";
```
connorbo97: test

src/components/Editor/components/Main.js
connorbo97: For now I think this is fine, but we should change this to check if there was an error/it failed to save. Maybe we can make the button red and make the text 'Failed to save'

src/components/Editor/components/Main.js
connorbo97: Don't think this style affects anything. When I ctrl-f there's only one instance of it. There was an earlier change that removed a lot of the inline styling so it was probably a mis-merge

src/components/Editor/components/SaveButton.js
connorbo97: Bc we're making a common Button later, no need to change, but in the future remember to update the component name!

PR #7, opened 2019-02-28
Change panel size to absolute value, changed panel bullets to buttons, add banner to disabled buttons, cleanup img folders and scraps file

connorbo97:
- Panel size is now controlled by a constant
- Bullets are now buttons with icons
- Buttons that aren't impelemented yet have banners to imply they are disabled versus a different styling
- Removed some extra images in img folder and random scrap file that shouldn't have been committed


src/styles/Panel.css
```diff
@@ -1,82 +1,148 @@
...
+  justify-content: flex-start;
+}
+
+.panel-options-list {
+  display: inline-block;
+  padding: 0px;
+  margin: 35px 25px;
+}
+
+/* .panel-options-item {
```
malsf21: Is there a reason that these lines are commented out? Will we use this styling in the future?

src/styles/Panel.css
```diff
@@ -1,82 +1,148 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  color: white;
+  position: relative;
+  margin-bottom: 20px;
+}
+.panel-item-content {
+  position: relative;
+  font-size: 33px;
```
malsf21: should we hard code w/ pixels or use em for font sizes, thoughts @connorbo97

connorbo97: Still unsure of the px vs em relationship and when to use em tbh. Can you explain the main reason to switch?

src/styles/Panel.css
connorbo97: Those are the stylings for the bullet. It was mroe just in case the buttons didn't end up working out. I can remove em

PR #9, opened 2019-03-07
Pretty error

alyssamw:


src/components/app.js
```diff
@@ -46,12 +46,14 @@ class App extends React.Component {
...
       } else {
-        this.setState({ checkedAuth: true, errorMsg: "No UID provided with user" });
+        this.props.loadFailure();
+        this.setState({ checkedAuth: true });
       }
     } else {
       console.log("no user found");
       this.props.clearUserData();
-      this.setState({ checkedAuth: true, errorMsg: "" });
+      this.props.loadFailure("No user could be found");
```
connorbo97: This else isn't a bad thing, it just means that no one is logged into firebase yet. So no need to create an error here!

src/reducers/userDataReducers.js
```diff
@@ -27,8 +27,7 @@ function userDataReducers(state = initialState, action) {
     case CLEAR_USER_DATA:
       return initialState;
     case LOAD_FAILURE:
-      state.error = "Failed to load user data...";
-      return state;
+      return Object.assign({}, state, { error: "Failed to load user data..." });
```
connorbo97: suggestion: have the error message be customizable for future use. Cuse now all errors will say failed to load user data

connorbo97: So that means you'll have to update the calls in app.js to pass in a string

src/components/app.js
```diff
@@ -105,6 +107,17 @@ class App extends React.Component {
             path="/createUser"
             render={() => (isValidUser ? <Redirect to="/editor" /> : <CreateUserPage />)}
           />
+          {/* Default error page */}
+          <Route
+            path="/error"
+            render={() =>
+              this.props.errorMsg ? (
```
connorbo97: Bc of the error created above in onAuthHandler, if you're not logged in and go to /error and refresh, you'll remain on /error

connorbo97:
Also, update the error message for when the server isn't working. Currently the error is "CATCH_ERROR". Located at getUserData in fetch.js, approx line 33

src/components/app.js
alyssamw: Should I just resolve this by deleting the path/error entirely? Am now unsure why in onAuthHandler, loadUserData is called with passing in showErrorPage.

connorbo97: loadUserData is passed showErrorPage cuse the 2nd param is a function to be called if an error occurs

AppContainer.js
```
 loadUserData: async (uid, onFailure) => {
      const { ok, data, error } = await fetch.getUserData(uid, true);
      //if the request went fine, and there's a non empty userData
      if (ok && data && data.userData && Object.keys(data.userData).length) {
        dispatch(loadUserData(uid, data.userData));
        if (data.programs) {
          dispatch(loadPrograms(data.programs));
        }
      } else {
        onFailure(error);
      }
    },
```

PR #14, opened 2019-03-07
Change code color scheme

connorbo97:
- try out some new colors to improve readability on lighter screens


connorbo97:
The original design had it in this monocolor purple layout which I think looks more inviting than other editors, but it really does fail the readability test on some screens...

PR #16, opened 2019-03-09
add ui reducer and update components to use new reducer state

connorbo97:
- Add uiReducer and actions
- Remove all component implemented event listeners for resize and replaced them with uiReducer state


malsf21:
On first glance, this looks like clean code. I'll do another more thorough passover sometime this weekend.

PR #19, opened 2019-03-26
add scripts for production builds using environment variables

connorbo97:
- Adds environment variables that trigger local vs production build


malsf21:
Looks like this is good from the Travis PR check passing, but do you want me to manually test this anyways? And w/ what env variables?

PR #20, opened 2019-03-27
Fixing Up Save Button + Disabling Save-On-Exit

malsf21:
This PR does three things:
* Fixes the implementation of the save button to pass on the correct JSON object to the endpoint, so the save button now works as intended
* Adds a "Saving..." indicator to the button while saving is going on - switches to "Saved!" once the fetch is completed
* Comments out `this.checkDirty()` from `TextEditor.js`, which disables the save-on-exit feature - which was causing memory leaks


malsf21:
Thanks for the fix, @connorbo97

PR #21, opened 2019-03-29
Makes Login & Create User Pages Responsive

malsf21:
This PR makes the Login and Create User pages more responsive - the forms now resize properly based on the viewport width, and the illustration scales appropriately and disappears when the viewport is too small.

In addition, I clean up a lot of the CSS & HTML behind the pages - but there's a lot more work to be done, probably in a refactor that creates just one Login component (opposed to the Modal & the Input), cleans up the styles even further into reusable components (which could involve a lot of renaming), and makes the Create User page independent of Login's styling (which is technically bad practice).

### DIFF
--- a/src/components/CreateUser.js
+++ b/src/components/CreateUser.js
@@ -23,23 +23,6 @@ class CreateUser extends React.Component {
     };
   }
 
-  renderModal = () => {
-    return (
-      <div className="login-page-content">
-        <div style={{ height: "0px" }}>&nbsp;</div>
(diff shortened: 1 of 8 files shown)


connorbo97:
Still getting some overlap of the background image and the input fields. It's a tough issue though cuse it's a combination of height and width

image

image

malsf21:
Yeah, I think it's a bit janky because of how the image is done - "in theory", we should probably make it a columned image instead. I might try playing around with the background image properties and see if that helps.

PR #22, opened 2019-03-29
Make Editor responsive to page width

connorbo97:
- add logic to Editor Output and Main.js to render differently at a breakpoint
- add new Radio component to select different options

### DIFF
--- a/src/components/Editor.js
+++ b/src/components/Editor.js
@@ -65,7 +65,7 @@ class Editor extends React.Component {
   };
 
   render() {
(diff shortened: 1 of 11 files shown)


malsf21:
Hm, if I select Code or Output, and size down (but still wide enough to technically fit both), it'll switch to both - I feel like this is unintuitive behaviour, and it should just stay on Code or Output; the only time the program should force you to switch is if you're on both and you dip below the specified viewport width. Not the end of the world, since this is a small use case, but I think it'd be useful.

PR #26, opened 2019-04-10
Font Awesome added

utsavm9:
As instructed on [Font Awesome](https://fontawesome.com/how-to-use/on-the-web/using-with/react) page, the Font Awesome package has been added.

The Hamburger icon is in 🍔! It changes to panel collapse panel when the panel is wide open.
The panel close button inside the panel is removed.

### DIFF
--- a/src/components/Editor/components/Main.js
+++ b/src/components/Editor/components/Main.js
@@ -7,6 +7,7 @@ import EditorButton from "./EditorButton";
 import * as fetch from "../../../lib/fetch.js";
 import EditorRadio from "./EditorRadio.js";
 import { EDITOR_WIDTH_BREAKPOINT, CODE_AND_OUTPUT, CODE_ONLY, OUTPUT_ONLY } from "../constants";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 
 /**------Props-------
  * textEditorSize: number? representing the percentage of space the left split pane takes up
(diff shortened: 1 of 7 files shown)


src/styles/Editor.css
```diff
@@ -27,16 +27,16 @@
...
 
 .editor-expand-panel-arrow {
-  color: #dddcdf;
-  font-weight: bold;
+  color: rgb(39, 33, 52);
   padding: 1rem;
   cursor: pointer;
   flex: 0 1 auto;
-  width: 20px;
+  transition: 1s;
```
connorbo97: what does this transition do?

src/styles/Panel.css
```diff
@@ -13,18 +13,19 @@
 }
 
 .panel-collapse-button {
-  position: absolute;
-  top: 10px;
-  right: 10px;
   color: #dddcdf;
+  padding-right: 15px;
```
connorbo97: add flex: 0 0 auto; to ensure that this doesn't get squished by the flex box when we run out of space
image

src/components/Editor/components/Main.js
```diff
@@ -71,19 +72,15 @@ class Main extends React.Component {
...
+    //if the left panel is closed, show nothing
+    //otherwise show hamburger icon
     return (
       <div
         className="editor-expand-panel-arrow"
         title="Open Profile Panel"
         onClick={handleOnVisibleChange}
       >
-        >
+        {panelVisible ? "" : <FontAwesomeIcon icon="bars" />}
```
malsf21: I think that this icon should be white - it seems inconsistent with the current colour that it is.

malsf21:
@connorbo97  the idea is to use [tree-shaking](https://fontawesome.com/how-to-use/with-the-api/other/tree-shaking) - take a look at the article, though I'm not sure how much we'd like to rework / if we need to eject.

src/styles/Editor.css
utsavm9: Transition was originally added when the bars turned slowly into a cross. But now I have seen the result after removing the transition and since the transition is unnoticeable, I'll remove this.

src/styles/Panel.css
utsavm9: Added!

utsavm9:
Is tree-shaking enabled in the production build just by including Rollup.js? Till then, I'll use deep-imports to help reduce the size of Font Awesome package.

malsf21:
We don't need to add any more things to the build haha, I believe that we'll have generic optimization by just importing what we need with explicit importing.

utsavm9:
A took a look at the stable build and I feel it would be user-eye-friendly if on hovering overt the panel buttons, the cursor did not change to a edit-bar which is indistinguishable from the button background.

PR #35, opened 2019-04-19
Uses Font Awesome's Explicit Importing

malsf21:
Changes around how #26 was implemented to exclusively use FA5's explicit imports. Should help with subsetting and etc. in the future.


malsf21:
The resolved conflict seems reasonable!

PR #44, opened 2019-05-04
create redux reducers for classes

connorbo97:
- Create the redux architecture for the classes and class views

### DIFF
--- a/src/reducers/classesReducer.js
+++ b/src/reducers/classesReducer.js
@@ -1,0 +1,22 @@
+import { DELETE_CLASS, LOAD_CLASSES, CLEAR_CLASSES, ADD_CLASS } from "../actions/classesActions.js";
+import Immutable from "immutable";
+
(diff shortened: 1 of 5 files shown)


connorbo97:
Let's sync up with others first

src/reducers/classesReducer.js
```diff
@@ -0,0 +1,22 @@
...
+let initialState = Immutable.List();
+
+function classesReducer(state = initialState, action) {
+  switch (action.type) {
+    case LOAD_CLASSES:
+      return Immutable.fromJS(action.classes);
+    case ADD_CLASS:
+      return state.push(action.data);
+    case DELETE_CLASS:
+      //TODO: look up Immutable API on how to remove a nested key
```
malsf21: Just a note - we need to do this!

malsf21:
Looks like @willob99 added this independently, going to close for now.

PR #45, opened 2019-05-13
Adds Language Icon to Editor Pane's Program Dropdown

malsf21:
I've been procrastinating doing the sketches finishing touches, so instead I quickly added the language icon to the editor dropdown. @connorbo97 , do you think this is the most elegant way to do this? Not sure if there's a more Immutable-y way to do it. Let me know what you think!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1068,6 +1068,21 @@
         "@fortawesome/fontawesome-common-types": "^0.2.17"
       }
     },
+    "@fortawesome/free-brands-svg-icons": {
+      "version": "5.8.2",
+      "resolved": "https://registry.npmjs.org/@fortawesome/free-brands-svg-icons/-/free-brands-svg-icons-5.8.2.tgz",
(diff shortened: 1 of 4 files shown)


malsf21:
Also, as a side note - it breaks on Travis because of an unused variable in the `find` function in `DropdownButtonContainer.js` - this is because I need to specify (value, key) in the function, but I only need key (but need to specify value). Am I using the wrong function for this? Looked through the Immutable docs but couldn't find what I needed.

connorbo97:
Can't look rn, but if you have an unused param, you could use _ instead of giving it a name?

malsf21:
Mm good point, I'll give it a shot. No rush ofc, but lmk if you know a better function to use. The actual feature works as intended.

malsf21:
Ok, I changed it so that all the programs in the dropdown menu have icons, and the way I did it then removed all the jank. I think it's not bad now, what are your thoughts?

PR #48, opened 2019-05-22
add Confirm Password to createUser, and add sign out button to error …

connorbo97:
- add confirm password field to login page


connorbo97:
I personally like clearing bc you don't really know which password you messed up. We can bring up during the meeting whether we wanna add a show password button instead. If we do that, we wanna not clear it so you can show password if you wanna see where ya messed up

connorbo97:
We can wait to merge this till the meeting

PR #49, opened 2019-05-24
Finishes Sketches View

malsf21:
This PR has a lot of things in it:
* Creates more sub-components for the Sketches View (SketchBox, more modals)
* Makes SketchBox larger
* Fixes some sketchy CSS with how the modals and SketchBoxes are created
* Adds an edit and delete sketch modal and button functionality
* Adds edit and delete sketch reducers and actions, adds setProgramName, setProgramLanguage, setProgramThumbnail reducers
* Reworks Sketches to use new DB for mat of keys instead of names 
* Merges delete program fetch functionality and ports over DB format fixes from `overhaul-db-architecture`
* Makes console icon use FontAwesome

We need to merge `delete-program-endpoint` and other backend overhaul branches before this should be merged.

Note: in the future, we should remove some of the fallbacks on the old DB stuff, consider creating more subcomponents (e.g. making a thumbnail picker component)


src/components/Sketches/components/EditSketchModal.js
```diff
@@ -0,0 +1,323 @@
...
+    if (this.badNameInput() || this.badLanguageInput()) {
+      // note it's impossible to have a bad thumbnail input
+      return;
+    }
+
+    let data = {};
+    let doUpdate = false;
+
+    if (this.state.newLanguage !== -1) {
+      data["language"] = this.state.newLanguage.value;
```
connorbo97: is there a reason for the ["string"] notation? Or just not a fan of . notation?

src/components/Sketches/components/EditSketchModal.js
```diff
@@ -0,0 +1,323 @@
...
+        fetch
+          .updatePrograms(this.props.uid, updateData)
+          .then(res => {
+            return res.json();
+          })
+          .then(json => {
+            if (!json.ok) {
+              this.setState({
+                spinner: false,
+                error: json.error || "Failed to create sketch, please try again later",
```
connorbo97: Failed to **edit** sketch, please try again later

src/components/Sketches/components/EditSketchModal.js
```diff
@@ -0,0 +1,323 @@
...
+            }
+            if (this.state.newThumbnail !== -1) {
+              this.props.setProgramThumbnail(this.props.sketchKey, this.state.newThumbnail);
+            }
+            this.closeModal();
+          })
+          .catch(err => {
+            this.setState({
+              spinner: false,
+              error: "Failed to create sketch, please try again later",
```
connorbo97: Failed to **edit** sketch, please try again later

src/components/Sketches/components/EditSketchModal.js
```diff
@@ -0,0 +1,323 @@
...
+        console.log(err);
+      }
+      this.setState({ spinner: true, error: "" });
+    }
+  };
+
+  renderMainModal() {
+    let thumbnailPreview =
+      this.state.newThumbnail !== -1 ? (
+        <img
```
connorbo97: can you move the ternary into the src prop instead of having 2 img's? Improves readability IMO

src/components/Sketches/components/EditSketchModal.js
```diff
@@ -0,0 +1,323 @@
...
+        />
+      ) : (
+        <img
+          src={`${process.env.PUBLIC_URL}/img/sketch-thumbnails/${this.props.sketchImg}.svg`}
+          className={"sketches-modal-header-thumbnail"}
+          alt="icodn"
+        />
+      );
+    return (
+      <ReactModal
```
connorbo97: I know you're not a fan of the renderBlah functions, but imo, they improve comprehension of these big jsx's.

src/components/Sketches/components/EditSketchModal.js
```diff
@@ -0,0 +1,323 @@
...
+  }
+  componentWillMount() {}
+  componentDidUpdate() {}
+
+  closeModal = () => {
+    if (this.props.onClose && {}.toString.call(this.props.onClose) === "[object Function]") {
+      this.props.onClose();
+    }
+    this.setState({
+      next: false,
```
connorbo97: next state not used

src/components/Sketches/components/SketchBox.js
```diff
@@ -0,0 +1,44 @@
...
+
+import "../../../styles/Sketches.css";
+
+import { Row, Col } from "reactstrap";
+import { faEdit } from "@fortawesome/free-solid-svg-icons";
+import { faTrashAlt } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+
+class SketchBox extends React.Component {
+  render() {
```
connorbo97: Thoughts on stateless components? We have a couple in the app, but we should be consistent. I'm impartial to either

connorbo97: As in the 
```
const Sketchbox = props => {
  return (
     Render equivalent
  )
}
```
style

src/components/Sketches/index.js
```diff
@@ -77,30 +90,49 @@ class Sketches extends React.Component {
...
+            );
+          }}
+          redirFunc={() => {
+            this.redirectToEditor(key);
+          }}
+        />,
       );
     });
-
+    console.log(sketches);
```
connorbo97: remove log

malsf21: Fixed!

src/components/Sketches/components/EditSketchModal.js
malsf21: Fixed!

src/components/Sketches/components/EditSketchModal.js
malsf21: Fixed!

src/components/Sketches/components/EditSketchModal.js
malsf21: Fixed!

src/components/Sketches/components/EditSketchModal.js
malsf21: Fixed!

src/components/Sketches/components/EditSketchModal.js
malsf21: Nah, no reason. Swapped it over.

src/components/Sketches/components/CreateSketchModal.js
```diff
@@ -151,8 +147,9 @@ class CreateSketchModal extends React.Component {
             });
             return;
           }
-          this.props.addProgram(this.state.name, json.data || {});
-          this.props.setMostRecentProgram(this.state.name);
+          console.log(json);
```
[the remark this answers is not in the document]
malsf21: Fixed!

malsf21:
> Very small tweaks. Overall looks solid. Can we also add in the width height 100% on the html doc into this change as well?

Fixed most of the changes. I think we should make that a separate PR for the width/height 100% because there might be more CSS we'd need to change (or CSS we'd want to change).

src/components/Sketches/components/SketchBox.js
malsf21: Yeah, I think that'd be reasonable. I have no opinion on stateless vs. normal components, haven't really had to decide on either. Any pros/cons we should look out for?

src/components/Sketches/components/EditSketchModal.js
malsf21: Reasonable, what should I break up? Probably each of the inputs, right?

connorbo97:
It should either mimic the behavior in the other modals (do nothing) or update the other modals to do the same for consistency actually

src/components/Sketches/components/SketchBox.js
connorbo97:
```
However, as of React v16, there are no performance benefits from using stateless functional components over class components. 
```
oof. well there's our answer. It makes it easier if you want to add state later if you have a stateful component so lets stick with this

src/components/Sketches/components/EditSketchModal.js
connorbo97: We can have a big ol readability change where we make everything consistent. We can talk bout it next week. For now, it's fine

connorbo97:
image
Also, unsure if it's cuse of my screen resolution or something, but the delete button overflows to a second line

malsf21:
Yeah, I'll make the edit modal do nothing then. As for the delete button, that just had to do w/ the modal size. I can just make delete one word then, or I can make the modal larger - you have any preference?

connorbo97:
Larger is my preference

malsf21:
So for reference, the create sketch button also loads the spinner - it's why I added it to the Edit Sketch modal. However, I can remove both, since the visual effect is hard to center properly, and then I'd just disable the button. Alternatively, I could use flexbox (wow) to vertically and horizontally center the spinner within the column container.

connorbo97:
I think the best choice is to disable the button while the fetch is happening (not sure if the reactstrap button has a disable value but I assume it does)

malsf21:
Cool, I can do that then. Will update you when I finish it!

malsf21:
Implemented button disabling & made the modal slightly larger. Can't fully fix the text-wrapping stuff without some creative CSS. LMK what you think!

connorbo97:
image
There's an odd gap at the bottom of the selectable thumbnails, but I'm not too bothered by it. Good to merge IMO

malsf21:
Oh I think it's because the hr tag isn't showing up properly, I can look into that. Doesn't break the merge though.

PR #52, opened 2019-06-30
Adds message in Sketches view when user has no sketches

malsf21:
Now, it prompts the user to create a new sketch! I think in the future, we could add a nice illustration, vertically center it, or add some spicy text. For now, this'll do.

### DIFF
--- a/src/components/Sketches/index.js
+++ b/src/components/Sketches/index.js
@@ -86,7 +86,7 @@ class Sketches extends React.Component {
       return (
         <div>
           <div className="no-sketches-container">
-            <h2>There's nothing here! Why don't you try creating a sketch!</h2>
+            <h2>There's nothing here! Why don't you try creating a sketch?</h2>
             <p>
               <SketchesButton
                 handleClick={() => this.setCreateSketchModalOpen(true)}
--- a/src/components/Sketches/index.js
+++ b/src/components/Sketches/index.js
@@ -82,6 +82,22 @@ class Sketches extends React.Component {
 
   renderSketches = () => {
     let newList = this.props.programs.concat([]);
+    if (newList.size === 0) {
+      return (
+        <div>
+          <div className="no-sketches-container">
+            <h2>There's nothing here! Why don't you try creating a sketch!</h2>
+            <p>
+              <SketchesButton
+                handleClick={() => this.setCreateSketchModalOpen(true)}
+                text={"Create A Sketch"}
+                width={"200px"}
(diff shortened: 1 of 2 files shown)


src/components/Sketches/index.js
```diff
@@ -82,6 +82,22 @@ class Sketches extends React.Component {
 
   renderSketches = () => {
     let newList = this.props.programs.concat([]);
+    if (newList.size === 0) {
+      return (
+        <div>
+          <div className="no-sketches-container">
+            <h2>There's nothing here! Why don't you try creating a sketch!</h2>
+            <p>
+              <SketchesButton
```
connorbo97: Going to need to update this to the new button when you merge the style change

connorbo97: technically second sentence is a question but I'm nitpicking

connorbo97:
https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/Link.md

I think replacing all the hack-y redirects with this Link component should fix the issues.

There's also a warning for putting a div inside of a paragraph block, but it technically doesn't break anything.

connorbo97:
And fix the unique key issue in the panel ples, we keep forgetting to squeeze that into a change and now is the time lol

malsf21:
> And fix the unique key issue in the panel ples, we keep forgetting to squeeze that into a change and now is the time lol

I fixed this one in the style overhaul PR, so it should be resolved there.

malsf21:
> Weird bug I found. Delete all your sketches. Create a sketch using the button from this change. After you're redirected to the editor, try to navigate to sketches using the side panel. Doesn't work for me.
> 
> But, if you refresh, the issue resolves itself.

Yeah, so I encountered this bug and put it in the Trello. I'll try the Link component in another branch.

src/components/Sketches/index.js
malsf21: > Going to need to update this to the new button when you merge the style change

Can't believe I didn't do it in this PR. It's a bit annoying since I think I'm going to get a merge conflict off of importing button, so I'll probably do this in whichever of the two branches I end up merging last.

Gonna tag #56 just so I remember.

malsf21: > technically second sentence is a question but I'm nitpicking

Fixed.

malsf21:
> ?

I'll merge this right now and add the Link change after I fix #56 - I think it'd be easier to work with.

PR #53, opened 2019-07-13
Adds Download Sketch Button

malsf21:
This is a pretty small PR that adds a "Download Sketch" button to each sketch in the sketches page. Two quick things before we merge:
1. Do we want to have a download page on the editor too? If so, what scope is the download function going to have?
2. What's the correct extension for processing? I've seen both `.processing` and `.pde`

### DIFF
--- a/src/components/Sketches/index.js
+++ b/src/components/Sketches/index.js
@@ -36,6 +36,30 @@ class Sketches extends React.Component {
     return SketchThumbnailArray[Math.floor(Math.random() * SketchThumbnailArray.length)];
   };
 
+  downloadSketchCode = (name, language, code) => {
+    let extension = ".";
+    switch (language) {
+      case "python":
+        extension += "py";
+        break;
+      case "processing":
+        extension += "processing";
+        break;
+      case "html":
+        extension += "html";
+        break;
+      default:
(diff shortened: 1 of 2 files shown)


connorbo97:
We should export Processing as an HTML page probably. Just take the contents of the iframe and export it as a .html file.

src/components/Sketches/index.js
```diff
@@ -36,6 +36,30 @@ class Sketches extends React.Component {
...
+        extension += "processing";
+        break;
+      case "html":
+        extension += "html";
+        break;
+      default:
+        extension += "txt";
+    }
+    // taken from this: https://stackoverflow.com/questions/44656610/download-a-string-as-txt-file-in-react
+    const element = document.createElement("a");
```
connorbo97: fancy

connorbo97:
Was able to download files. Python files give warnings about being dangerous but I'm not surprised about that.

malsf21:
> We should export Processing as an HTML page probably. Just take the contents of the iframe and export it as a .html file.

Cool, didn't think about that. Ended up using the `Processing.js` in Output. Check it out and let me know what you think, not sure how jank this is haha

PR #54, opened 2019-07-14
Adds More Descriptive Firebase Auth Errors

malsf21:
I took a look at the [Firebase Auth Errors](https://firebase.google.com/docs/reference/js/firebase.auth.Error) and added more descriptive messages to the Create User and Login Form pages. At this point, I've encompassed most of the common errors (invalid username, account taken, password or username improperly formatted, incorrect password, network issues, improper configuration).

### DIFF
--- a/src/components/CreateUser/CreateUserForm.js
+++ b/src/components/CreateUser/CreateUserForm.js
@@ -124,11 +124,46 @@ export default class CreateUserForm extends React.Component {
       .auth()
       .createUserWithEmailAndPassword(email, passHash)
       .then(({ user }) => {})
-      .catch(error => {
-        console.log(error);
-        let newErr = error.message || "";
-        newErr = newErr.replace("email address", "username");
-        this.setState({ waiting: false, errorMessage: newErr || "failed to create user" });
+      .catch(err => {
+        console.log(err);
+        let newMsg = err.message;
+        switch (err.code) {
+          case "auth/invalid-email":
+            newMsg =
+              "Invalid username inputted. Usernames must only have alphanumeric characters plus !@#$%.";
+            break;
+          case "auth/email-already-in-use":
+            newMsg = "Username is taken; please use another one.";
+            break;
+          case "auth/user-not-found": // this shouldn't ever happen
+            newMsg = "No account found for username.";
+            break;
+          case "auth/wrong-password":
+            newMsg = "Invalid password provided.";
+            break;
+          case "auth/network-request-failed":
+            newMsg = "Network error - check your internet connection.";
+            break;
+          case "auth/app-deleted":
+          case "auth/app-not-authorized":
+          case "auth/argument-error":
+          case "auth/invalid-api-key":
+          case "auth/operation-not-allowed":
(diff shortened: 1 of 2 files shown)


src/components/CreateUser/CreateUserForm.js
```diff
@@ -124,11 +124,46 @@ export default class CreateUserForm extends React.Component {
       .auth()
       .createUserWithEmailAndPassword(email, passHash)
       .then(({ user }) => {})
-      .catch(error => {
-        console.log(error);
-        let newErr = error.message || "";
-        newErr = newErr.replace("email address", "username");
-        this.setState({ waiting: false, errorMessage: newErr || "failed to create user" });
+      .catch(err => {
+        console.log(err);
```
connorbo97: Up to you if you want to keep this log in. I thought it would be nice if teachers were able to relay this to us. Technically, we could set up something with the backend to be able to log this type of stuff but for another day.

src/components/CreateUser/CreateUserForm.js
```diff
@@ -124,11 +124,46 @@ export default class CreateUserForm extends React.Component {
...
+        let newMsg = err.message;
+        switch (err.code) {
+          case "auth/invalid-email":
+            newMsg =
+              "Invalid username inputted. Usernames must only have alphanumeric characters plus !@#$%.";
```
connorbo97: inputted isn't a word apparently, just remove it and have it start with "Invalid username: "

src/components/CreateUser/CreateUserForm.js
```diff
@@ -124,11 +124,46 @@ export default class CreateUserForm extends React.Component {
...
+          case "auth/invalid-user-token":
+          case "auth/user-disabled":
+          case "auth/user-token-expired":
+          case "auth/web-storage-unsupported":
+            newMsg = "Issue with user. Please contact administrator. Error: " + err.code;
+            break;
+          default:
+            newMsg = "Failed to create user: " + err.code;
+        }
+        this.setState({ waiting: false, errorMessage: newMsg || "failed to create user" });
```
connorbo97: capital F in failed

src/components/Login/LoginForm.js
```diff
@@ -43,17 +43,37 @@ export default class LoginModal extends React.Component {
           console.log(err);
           let newMsg = err.message;
           switch (err.code) {
+            case "auth/invalid-email":
+              newMsg =
+                "Invalid username inputted. Usernames must only have alphanumeric characters plus !@#$%.";
```
connorbo97: Technically shouldn't happen bc this is for creating a user but just in case, remove the inputted like above.

src/components/CreateUser/CreateUserForm.js
malsf21: Agreed, I'll keep them in for now and switch it to `console.error()` so it's not just a random console log. I do wonder how we'd log this stuff eventually - definitely something to consider if we ever rewrite the backend.

malsf21:
Made all the changes!

PR #57, opened 2019-07-27
Adds "True Redirect" On GitHub Pages URLs

malsf21:
As mentioned [here](https://github.com/rafrex/spa-github-pages), GitHub Pages doesn't properly handle application paths for SPAs (like ours) - this means that users can't bookmark our pages, click on links that link to our non-app root, and generally do intuitive behaviour that they'd expect out of the website.

Or, as they explain it:

> GitHub Pages doesn't natively support single page apps. When there is a fresh page load for a url like example.tld/foo, where /foo is a frontend route, the GitHub Pages server returns 404 because it knows nothing of /foo.

This PR fixes that problem by implementing the solution in the aforementioned - it adds a custom `404.html` that redirects the user to `index.html`, and then passes in the previously entered URL into React's routing system, which should lead the user to the correct result.

The only unfortunate thing about this PR is that we can't exactly test it reliably without committing to master/deploying to GH Pages, but it should be an unobtrusive change that we can easily revert.

### DIFF
--- a/public/404.html
+++ b/public/404.html
@@ -1,0 +1,38 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Teach LA | 404: Redirecting You...</title>
+    <script type="text/javascript">
+      // Single Page Apps for GitHub Pages
+      // https://github.com/rafrex/spa-github-pages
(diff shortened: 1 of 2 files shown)


connorbo97:
I'll take your word on this one haha, I've never worked with GH pages, but you know what you're doin.

PR #58, opened 2019-08-07
Uses <Link> instead of <Redirect> for (some) Rerouting

malsf21:
This addresses a bug brought up in PR #52 - namely, that navigating to a newly created sketch, then clicking "sketches" doesn't bring the user back to the sketches view.

This PR swaps out the state-based loading of the `<Redirect>` component with directly using the `<Link>` component on the button that triggers the redirect. Now, the behaviour described above works as intended - this PR meets the bare minimum of fixing the bug.

Note that right now, this is only implemented on the Profile Panel's buttons (Editor, Sketches) and the individual Sketch Box button (that redirects to the editor) - this is because these two are guaranteed switches, while two other uses of `<Redirect>` (redirecting to sketches if the editor is empty, and the create sketch button only redirecting after the sketch has been successfully created) are conditional and cannot be replaced with `<Link>` by themselves. However, I'm open to any suggestions.

### DIFF
--- a/src/components/Sketches/components/SketchBox.js
+++ b/src/components/Sketches/components/SketchBox.js
@@ -1,5 +1,5 @@
 import React from "react";
-
+import { Link } from "react-router-dom";
 import "../../../styles/Sketches.css";
 
 import { Row, Col } from "reactstrap";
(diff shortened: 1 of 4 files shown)


malsf21:
Going to force merge this since it's a significant bugfix. Sorry in advance 😢

PR #60, opened 2019-09-28
Changes "Run" button's icon to play arrow

malsf21:
As the title says, all this PR does is change the "Run" button's icon to a play arrow (from a checkmark), since that's more intuitive with "running" a program. Also leaves space to use a pause icon in the future if needed.


malsf21:
thanks 😄

PR #90, opened 2019-10-23
Adds .nvmrc file

malsf21:
This is a very small PR, it enforces node `v10.15.3` via a `.nvmrc` file (which only kicks into action if the user has `nvm` installed). Should just make life easier for our new developers!


malsf21:
Going to force merge this - it literally can't cause any problems.

PR #94, opened 2019-10-24
Reverts react-redux to 5.1.2

malsf21:
Turns out that we weren't good to merge react-redux `7.*` into our codebase since it causes an issue with `react-redux-firebase`, and I wasn't able to resolve it quickly. Merging this until we can deal with the problem.


malsf21:
Going to merge this as admin unless it doesn't pass CI.

PR #95, opened 2019-10-24
Fixes Cursor Position for Chrome

malsf21:
Chrome has updated their `::after` to be similar to Firefox's behavior (#55), so this updates the base default to use the "new" behavior, and adds a Safari exception. Needs more field testing.


malsf21:
(going to force merge as well)

PR #98, opened 2019-10-26
Turn "Save" and "Run" into icon buttons when screen gets too small

krashanoff:
This is a super tiny PR but here's what it includes:
* Add `.editor-button-text` class. It is currently applied to the "Save" and "Run" labels for the editor. There are no custom styles for it as of this writing.
* Conditionally render "Save" and "Run" labels when the screen is not "too small"
    - "Too small" means that the viewport meets the requirement for `isSmall`: the screen width is less than or equal to the `EDITOR_WIDTH_BREAKPOINT` constant in `Editor/constants/index.js`.

### DIFF
--- a/src/components/Editor/components/Output.js
+++ b/src/components/Editor/components/Output.js
@@ -158,7 +158,7 @@ class Output extends React.Component {
       {this.renderConsoleButton()}
       <Button className="mx-2" color="primary" size="lg" onClick={this.runCode}>
         <FontAwesomeIcon icon={faPlay} />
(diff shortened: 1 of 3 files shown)


malsf21:
Also, a quick note - noticed a bug you get when you start in "Output" on a small viewport and expand to a larger viewport.

malsf21:
And, another note - I'd eventually like to test this with Enzyme/Jest. Most likely will check CSS styling based on viewport width.

PR #103, opened 2019-10-30
[DRAFT] Implements View-Only Sketch Page

malsf21:
Draft PR, will get updated as more info gets added.

So far, added the implementation for `getSketch(docID)`,  and the router path - will need to implement view (which @emmyc is taking the lead on atm).


src/components/app.js
```diff
@@ -65,6 +66,11 @@ class App extends React.Component {
     }
   };
 
+  getProgram = async programid => {
+    const { ok, sketch } = await fetch.getSketch(programid);
+    return ok ? true : false;
+  };
+
```
malsf21: @alyssamw this breaks the CI, not using `sketch`

src/components/app.js
```diff
@@ -65,6 +66,11 @@ class App extends React.Component {
     }
   };
 
+  getProgram = async programid => {
+    const { ok, sketch } = await fetch.getSketch(programid);
+    return ok && sketch ? true : false;
+  };
+
```
malsf21: Does this pass the sketch information to the routed path? Just noticed that you're not actually returning any sketch information to `MainContainer`.

alyssamw: For now it's not passed. Should I just get rid of it or might we need the sketch info later?

malsf21: Right right, I think it'll get passed to the component - that's what I meant by dummy component earlier. Let me see if I can patch together a basic version for this branch.

As a sidenote, means we're probably not going to merge until the feature is done.

malsf21:
Going to merge `master` into this now - it'll be the final branch where we add the actual functionality for `view-only`.

malsf21:
Officially tagged @emmyc into this. Will likely merge our theming PR into this branch, since we'll use that for the view-only view.

malsf21:
Remerged master into this branch. Officially re-tagging @jamieliu386 back into the ticket.

malsf21:
Superseded by #219

PR #104, opened 2019-10-30
Updates skulpt to v1.1.0

malsf21:
After receiving some feedback from the instructors, bumping up the skulpt version might improve some reliability we'll get out of the Python "interpreter". 

Two other quick notes: we should support both Python 2 and Python 3, and this entire component is super jank.


malsf21:
Yeah, there are quite a few different JS implementations of Python - it's something we'll explore a little down the line.

PR #105, opened 2019-10-30
Button group visual changes

emmyc:
Changed colors of view toggle button group to purples to match theme more, also adjusted border radius and padding

### DIFF
--- a/src/styles/Radio.scss
+++ b/src/styles/Radio.scss
@@ -1,3 +1,5 @@
+@import "styles/variables.scss";
+
 .radio-selector {
   display: flex;
   height: 60px;
@@ -5,8 +7,7 @@
 }
 
 .radio-option {
-  //   border: 2px solid #ddd;
-  background-color: #473d5d;
+  background-color: $theme-light;
   color: #eee;
   padding: 8px;
   cursor: pointer;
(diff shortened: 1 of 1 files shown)


src/styles/Radio.scss
```diff
@@ -1,41 +1,52 @@
...
+    border-left: none;
+  }
+}
+.radio-option-selected {
+  border: none !important;
+  background-color: #7a45f0;
+  color: #fff;
+  padding: 8px;
+  cursor: pointer;
 }
```
malsf21: Why is this duplicated twice?

src/styles/Radio.scss
```diff
@@ -1,41 +1,52 @@
...
 
 .radio-option {
-    border: 2px solid #ddd;
-    background-color: rgb(238, 170, 25);
-    color: #eee;
-    padding: 5px;
-    cursor: pointer;
-    line-height: 27.5px;
+  //   border: 2px solid #ddd;
+  background-color: #473d5d;
```
malsf21: This is the same as the SASS variable `$theme-light`, can you change it to that

src/styles/Radio.scss
```diff
@@ -1,41 +1,52 @@
...
 
 .radio-option-selected {
-    border: 2px solid #ddd;
-    background-color: #4caf50;
-    color: #eee;
-    padding: 5px;
-    cursor: pointer;
-    line-height: 27.5px;
+  border: none !important;
+  background-color: #7a45f0;
```
malsf21: Not a required change, but I think it might be good if we could change this to a SASS variable. Eventually we'll be writing some custom themes so variables make our life easier!

connorbo97:
Disagree, the pill is perfect.

PR #106, opened 2019-11-01
Resizes Default Ratio of Code:Output to be 3:2

malsf21:
This is a quick PR that changes the default ratio of Code : Output components on the Editor to be 3:2, which gives us more breathing space in the top of the code editing pane.


krashanoff:
Will review this evening. Sorry for the delay.

PR #108, opened 2019-11-02
Updates Unsafe Lifecycles

malsf21:
This is a PR that updates the "UNSAFE" lifecycles mentioned in [this blog post](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html), namely for us `componentWillMount`. Mostly, this PR just removes stubs of functions that were never used, but it also changes `componentWillMount` to `componentDidMount` in `app.js` and `Editor/index.js`; the event listeners still function as intended.

As a note, we'll also have to greenlight the new `react-router-dom` version to fully remove all of the warnings.

### DIFF
--- a/src/components/Editor/components/TextEditor.js
+++ b/src/components/Editor/components/TextEditor.js
@@ -31,8 +31,6 @@ class TextEditor extends React.Component {
     window.addEventListener("close", this.onLeave);
   }
 
-  componentWillUpdate() {}
-
   componentWillUnmount = () => {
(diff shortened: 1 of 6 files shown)


malsf21:
> Untested code, but blogpost recommends just moving initial state to constructor, so we don't need componentDidUpdate

Tested locally, change works. You're good to greenlight the PR.

In general though, I'd caution against tagging-on other changes to PRs. Worked out for this though!

PR #109, opened 2019-11-04
Added support for Redux Devtools Extension

fan-matt:
Super small PR, just allows us to use the Redux browser extension to help us see how our state is changing

https://extension.remotedev.io/


fan-matt:
> Looks good to me, doesn't break CI. Are you using the chrome extension?

I'm personally using firefox but chrome should be fine

PR #110, opened 2019-11-04
Fixes incorrect usage of <Link> in Error component

malsf21:
Previously, the `Error` component could get returned without a wrapping `<Router>` and/or `<Switch>`, which violates `react-router-dom`'s rules and makes it throw a very cryptic error. This PR fixes that, by making all calls to the `Error` component properly wrapped in an accompanying router. It also makes the `Error` component just a bit clearer.

(To reproduce the bug, force an error message app-wide, like having a wrong server URL)

### DIFF
--- a/src/components/Error.js
+++ b/src/components/Error.js
@@ -4,28 +4,30 @@ import Footer from "./common/Footer";
 import { Link } from "react-router-dom";
 import firebase from "firebase";
 import "styles/Page.scss";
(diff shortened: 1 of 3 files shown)


malsf21:
> I wish there was a slightly cleaner way of going about this.

Yeah, I agree. Dislike the lack of a true global router, but not too much we can do right now (I think, I'm no react-router expert?).

PR #112, opened 2019-11-04
[DRAFT] Prep for Netlify Deployment & App at Root of Domain

malsf21:
I love GitHub Pages, but it's not super flexible (and the whack 404 is never something I love). We're going to transition to Netlify, where we can easily manage staging, PR previews, etc. This PR does the first set of legwork needed to make that change (putting the app at the root of the domain).

* changes `segmentCount` in `public/404.html` to `0`, since we're now rooted off of the domain
* in `constants/index.js`, changes react router's base URL to just `/`
* comments out GH build process in Travis - might reintroduce for staging if it's trickier to do in Netlify
* changes `homepage` in `package.json` to match new domain-rooted app

Note: blocked by DNS changes ~& `fix-link-outside-router`~


malsf21:
Hm, we actually need the `404.html`, even for Netlify. Good to know!

malsf21:
Going to force merge now, Howard has CNAMEd the domain!

PR #114, opened 2019-11-05
Add download code button to main editor

fan-matt:
Adds this big nifty green button that downloads your code! Spiffy.

Also acts as a staging ground for future refactoring and code organization by splitting common download logic to downloader utility.

download button

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -9,6 +9,7 @@ import * as fetch from "../../lib/fetch.js";
 import EditorRadio from "./components/EditorRadio.js";
 import { Redirect } from "react-router-dom";
 import { EDITOR_WIDTH_BREAKPOINT, CODE_AND_OUTPUT, CODE_ONLY, OUTPUT_ONLY } from "./constants";
+import CodeDownloader from "./../../util/languages/CodeDownloader";
 
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faSave } from "@fortawesome/free-solid-svg-icons";
@@ -86,6 +87,10 @@ class Editor extends React.Component {
     this.props.cleanCode(this.props.mostRecentProgram); // Set code's "dirty" state to false
   };
 
+  handleDownload = () => {
+    new CodeDownloader().download(this.props.name, this.props.language, this.props.code);
+  };
+
   renderDropdown = () => <DropdownButtonContainer />;
 
   renderCodeAndOutput = () => (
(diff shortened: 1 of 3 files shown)


src/components/Editor/index.js
```diff
@@ -141,6 +147,10 @@ class Editor extends React.Component {
             <span className="editor-button-text">&nbsp;&nbsp;{this.state.saveText}</span>
           )}
         </Button>
+
+        <Button className="mx-2" color="success" size="lg" onClick={this.handleDownload}>
+          <FontAwesomeIcon icon={faDownload} />
+        </Button>
```
malsf21: I think the spacing for the button is a bit off (the x-margin is larger between this and the save code option compared to the save code button and the pill) - could you fix that?

src/util/languages/CodeDownloader.js
```diff
@@ -0,0 +1,30 @@
...
+      file = new Blob([ProcessingConstructor(code, true)], { type: "text/plain" });
+    } else {
+      file = new Blob([code], { type: "text/plain" });
+    }
+    element.href = URL.createObjectURL(file);
+    element.download = name + extension;
+    document.body.appendChild(element); // Required for this to work in FireFox
+    element.click();
+  };
+}
```
malsf21: This might be a bit of bike-shedding, but should this explicitly be a class? To me, this feels more of a single helper-type function that can be exported from a file.

Then, the API becomes something like `helpers.downloadCode(...)` rather than `new CodeDownloader().download(...)`, since we don't even do anything with the `CodeDownloader()` afterwards.

fan-matt: There's some discussion to be had here. I don't really think there's anything wrong with creating tiny classes- they modularize and make code a bit more explicit. I did go ahead and make the method static, so now it looks like `CodeDownloader.download(...)` . Let me know what you think- and of course, much of this is subject to change.

src/components/Editor/index.js
fan-matt: Would you rather me adjust the margins for the radio button? That would make it more consistent with the buttons in the output pane

fan-matt: I went ahead and did that: here is the result

buttons


It's quite difficult to change the margins on those buttons- they have a big fat `!important` next to them

malsf21: Hm, it still feels a bit off - but I think this is good for now. Looks like we might have to add another button in a moment so we'll see how this all works out. Solid work!

src/util/languages/CodeDownloader.js
malsf21: Hm okay, I see where you're coming from. LGTM off of functionality, I think we'll eventually need to formalize our conventions on helpers and such.

PR #116, opened 2019-11-08
FE: move images off of imgur

evazhog:
Downloaded the images off imgur to `public/img/icons` and linked to local directory on `src/constants/index.js`


src/constants/index.js
```diff
@@ -32,16 +32,17 @@ SUPPORTED_LANGUAGES.forEach(lang => {
   }
 });
 
+const public_url = process.env.PUBLIC_URL;
```
malsf21: This seems super small, but see if you can make constants upper-case - it's a common programming habit!

PR #118, opened 2019-11-09
Add ViewportAwareButton to Editor.

krashanoff:
Hey there. Remember on Friday when we were trying to figure out how to test my fix from earlier? Well, I made this to make the testing easier. This pull request adds the following:

* `ViewportAwareButton`
    - Works just like a reactstrap `Button`, but removes any children marked with the class `viewport-aware` when given the prop `isSmall`.
* Tests for `ViewportAwareButton`
    - With the above out of the way, testing is rather straightforward.

### DIFF
--- a/src/components/Editor/components/ViewportAwareButton.js
+++ b/src/components/Editor/components/ViewportAwareButton.js
@@ -1,0 +1,19 @@
+import React from "react";
+import { Button } from "reactstrap";
+
+/**-------Props--------
+ * isSmall: dictates display of anything marked with className '.viewport-aware'
+ * ...    : any props you would pass to a reactstrap button
+ */
+const ViewportAwareButton = props => {
+  // eliminate viewport aware children if small.
+  let { children, ...remainder } = props;
+  if (props.isSmall)
+    children = Array.from(props.children).filter(
+      child => child.props.className !== "viewport-aware",
+    );
+
+  return <Button children={children} {...remainder} />;
+};
+
+export default ViewportAwareButton;
--- a/src/components/Editor/components/ViewportAwareButton.js
+++ b/src/components/Editor/components/ViewportAwareButton.js
@@ -10,7 +10,7 @@ const ViewportAwareButton = props => {
   let { children, ...remainder } = props;
   if (props.isSmall)
     children = Array.from(props.children).filter(
-      child => child.props.className !== "viewport-aware",
+      child => !child.props.className.split(" ").includes("viewport-aware"),
     );
 
   return <Button children={children} {...remainder} />;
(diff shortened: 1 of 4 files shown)


krashanoff:
The original intent was to provide a sort of "drop-in replacement" for the `Button`s we were using. I see the appeal of making it a self-closing tag. Will make the suggested changes now.

src/components/Editor/components/ViewportAwareButton.js
```diff
@@ -0,0 +1,26 @@
...
+ * any props you would apply to a reactstrap Button.
+ */
+const ViewportAwareButton = props => {
+  // extract icon and text from props.
+  let { icon, text, ...remainder } = props;
+
+  // render icon always, text only if not small.
+  return (
+    <Button {...remainder}>
+      {icon}
```
malsf21: Sorry to be super nit-picky here, but I don't think you should pass down the entire `<FontAwesomeIcon>` as the prop for `icon`; the `ViewportAwareButton` should be the one that handles importing `<FontAwesomeIcon>` and inserting the resulting icon in this file. I think this makes the code a little clearer and slightly easier to use, at maybe the expense of future component flexibility.

@fan-matt any thoughts on this?

fan-matt: Here's my 2 cents on the issue:

I think you shouldn't pass in any children in as props since this makes the button very limited in what it looks like. To maximize our "drop-in replacement" theme we're going for, I think it makes sense to do something like this:

```javascript
// App.js
import React from 'react';

import Container from './Container';


function App() {
  return (
    <div className="App">
      <Container>
        <h1 className="a">I'm A</h1>
        <h1 className="b">I'm B</h1>
      </Container>
    </div>
  );
}

export default App;


// Container.js
import React from "react";

export default class Container extends React.Component {

    render() {
        console.log(this.props.children);

        let renderMe = this.props.children.filter( (child) => {
            return child.props.className === "a";
        });

        return(
            <div>
                {renderMe}
            </div>
        );
    }
}
```
This only renders "I'm A"

krashanoff: > it definitely sits better with me to write it like the previous iteration

+1 on this. I am flexible in how to implement this, but I thought the first iteration made a little more sense since one could use it as a "drop-in" replacement. It isn't necessarily a requirement that every `<ViewportAwareButton>` have an icon, just like how it isn't a requirement that every `<Button>` have one. `<Button>`s have bodies and children, as should their ViewportAware counterparts. However, it is a possibility that one would still desire to make certain components within said Button vanish when the screen gets too small, in which case a ViewportAware flavor is preferable.

> passing down the entire `<FontAwesomeIcon>` since it future-proofs this component

That was the logic in my decision there. We are using FontAwesome for the time being, but this is something that could change later down the line - hence passing the entire component as a prop. While I too am not a fan of passing the `icon` prop down this way, I figured it would be best for future-proofing as Matt mentioned.

malsf21:
Solid discussion, and I agree. Going to merge!

PR #130, opened 2019-11-20
Fixes URL/Resource for Lightbulb Profile Image

malsf21:
Uh oh, this was never done properly - but also my bad becuase we never caught this in testing.

Tagging @evazhog.


malsf21:
Force merging.

PR #134, opened 2019-11-25
Adds Travis Jobs to test for Node LTS, Node Latest

malsf21:
This is mostly a developer convenience PR - just adds two jobs that can fail to test for Node LTS and Node latest changes. Gives us a heads up for future feature implementation!


malsf21:
Looks like this all works, going to force merge.

PR #172, opened 2020-01-17
Modify Profile Panel's Name-Edit component to be better.

kilometerskang:
Changed cursor on hover.
Made font/padding consistent when editing name.
Added green check mark when name is submitted.
Removed blue outline from input form.


malsf21:
Closed by #174

PR #197, opened 2020-02-08
Added help text to loading page

rzlien:
Added help text to loading page if loading takes more than 2 seconds. Allows users to submit issues to Github through a link on the loading page.

### DIFF
--- a/src/components/common/LoadingPage.js
+++ b/src/components/common/LoadingPage.js
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React from "react";
 import { RingLoader } from "react-spinners";
 import "styles/Loading.scss";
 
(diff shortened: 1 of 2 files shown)


src/components/common/LoadingPage.js
```diff
@@ -10,11 +10,38 @@ import "styles/Loading.scss";
...
+    clearTimeout(this.setState.timer);
+  };
+
+  render = () => {
+    return (
+      <div className="Loading">
+        <div className="Loading-title">Loading</div>
+        <RingLoader color={"#171124"} size={250} loading={true} />
+        {this.state.showText && (
+          <p className="Loading-page-text" style={{ color: "white" }}>
```
malsf21: Right now, the `Loading-page-text` class isn't defined. However, you could add it (in `src/styles/Loading.scss`) and then apply CSS styles that way, instead of using the style tag.

Also, right now you have `color: white`, but we actually don't use white on the loading page, we use an off-white. In the `.scss` file, you can set `color: $off-white` (similar to the line for `.Loading-title`).

I'd also maybe add a bit of margin/padding to the top and the sides: it's a bit too close to the spinning loader, and if you shrink the viewport (i.e. make the browser smaller) the text sticks a bit too close to the sides.

src/components/common/LoadingPage.js
```diff
@@ -10,11 +10,38 @@ import "styles/Loading.scss";
...
+            Looks like loading is taking a bit long! If it takes too long, submit an issue on
+            <a href="https://github.com/uclaacm/TeachLAFrontend/issues"> Github</a>.
```
malsf21: This link doesn't contrast too well with the purple background. I'd probably create a class called `.Loading-link-text` (or something like that), and use

```
.Loading-link-text {
  color: $off-white;
  text-decoration: underline;
  ...
}
```

Which would make it contrast better with the background but still make it obvious that it's a link. You might also want to take a look at the `:hover` attribute, which will also need some styling.

PR #199, opened 2020-02-12
Adds link to GitHub Repo on 404 page, makes styling respect current theme

malsf21:
This is a pretty small and quick PR. Adds a link to our GitHub repo on our 404 page (which might happen eventually with the view-only programs), and makes the page styling respect the current theme (with the `themify` mixin). Also tidies up a bit of the underlying React code.

(note: this should be squash merged, commit history is not too useful).


malsf21:
> Hey Matt, looks good, but when you mention the styling should respect the current theme, I only ever get the light themed error page.

Good catch! The theme itself wasn't actually working, so that makes lots of sense. Just added on a commit that should properly respect the theme.

(in this case, I'm fine with pulling from `cookies` since getting nothing/invalid values has sensible defaults)

PR #210, opened 2020-02-21
Adds Tests for <Loading>'s Timer + Minor Improvements

malsf21:
This is a quick follow up to @rzlien's PR (#197) adding a help text timer to `<Loading>`. It does three things:

1. renames the prop `showText` to `showHelpText`, and makes it inherit the value from an optional prop
2. uses `GH_REPO_NAME` constant instead of hard link for repo
3. adds tests to check for `showHelpText` prop validity and to check that the timer works (i.e. on mount, the help text is not there, and then after 2s, the help text shows up)

Not super happy with how the timer test is written, since I wanted to use Jest's [mock timer calls](https://jestjs.io/docs/en/timer-mocks) - but I wasn't able to get them to work.

### DIFF
--- a/src/components/common/__snapshots__/LoadingPage.test.js.snap
+++ b/src/components/common/__snapshots__/LoadingPage.test.js.snap
@@ -1,5 +1,23 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`LoadingPage should handle showHelpText prop properly 1`] = `
+<div
+  className="Loading"
+>
+  <div
+    className="Loading-title"
+  >
+    Loading
+  </div>
+  <Loader
+    color="#171124"
+    css=""
+    loading={true}
+    size={250}
+  />
+</div>
+`;
+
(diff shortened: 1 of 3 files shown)


src/components/common/__snapshots__/LoadingPage.test.js.snap
```diff
@@ -1,5 +1,23 @@
...
+  <div
+    className="Loading-title"
+  >
+    Loading
+  </div>
+  <Loader
+    color="#171124"
+    css=""
+    loading={true}
+    size={250}
```
krashanoff: Tests failing on my system due to the snapshot:

image

malsf21: When's the last time you've updated your `node_modules`? I believe that we updated `react-spinners` which changes the props that `<Loader>` exposes. The test passes on both my local computer and the CI, so not exactly sure what's happening?

PR #211, opened 2020-02-22
Tries Out GitHub Actions for CI

malsf21:
For now, this just builds and tests our project on push - similar to our current CI workflow.


malsf21:
Going to force merge this since it's a DevOps change only.

PR #219, opened 2020-03-02
View-Only Sketches and Major Main/Editor Refactoring

malsf21:
Wow, it's a relief to finally write out this PR! In sum, this is the PR that implements all the necessary work for a View-Only page for each sketch (which works whether or not the user is logged in). It required a lot of code motion and refactoring, and likely has lots of side effects (so I'd prefer some semi-rigorous testing and multiple reviewers). Many thanks to @jamieliu386 and @emmyc for the solid (and gruelling) work they put into this feature!

(this supersedes #103)

Here's a semi-detailed list of what we did:
* moves the logic that combines `Editor`, `Output`, and `SplitPane` out of `Main` to new component `EditorAndOutput`; similarly moves requisite props/state and `CodeDownloader` work. necessary to remove code duplication
* creates `ViewOnly` and accompanying `ViewOnlyContainer` for view-only programs. this component manages fetching the program from the pid, handling errors, and passing down the relevant props to `EditorAndOutput`
* creates app route of `/p/:pid` for users to visit view-only sketches - directs to `ViewOnly`
* adds custom props `this.props.viewOnly`, `this.props.vLanguage` to `Output`, to process props differently depending on program state
* makes `Editor` banner behaviour (e.g. Dropdown vs. text, save button) dependent on view-only mode
* adjusts `ProfilePanel` behaviour to have Create User and Login buttons when the user isn't logged in; maintains themeing capabilities
* adjusts tests for profile panel for new behaviour (TODO: test non-logged in profile panel)

(also, we should probably squash merge this, but I'm not sure how useful the commit history actually might be!)


malsf21:
Just for transparency, @jamieliu386 discovered a bug where we corrupt the Redux state while viewing a view-only program when logged-in. Currently working on an (elegant) fix right now!

malsf21:
> Finished parsing through the last few files and played around with the view only editor more. The only thing I stumbled on was that the view only editor does not load the set theme from cookies an user's cookies

Ah, that seems like a bug. If you have the time to take a look at that, that'd be great, but if not, I can probably fix it tomorrow!

malsf21:
Jeez, I think this is going to live in some sort of infamy.

PR #225, opened 2020-03-09
Removing Defunct Code

malsf21:
We won't be supporting C++/Java any time soon, and we no longer use the Facebook Auth provider. Removes code references to all of these things.


vohndernet:
Looks great! I give the all clear

PR #226, opened 2020-03-09
Removes Travis

malsf21:
Looking to switch to GitHub Actions. Blocked until further discussion.


malsf21:
I've had "further discussion" (with myself haha), I think we're good to move off of Travis for now.

PR #235, opened 2020-03-20
Switch to Go backend; modify backend request pattern.

krashanoff:
Ahoy! This is a PR to switch out our backends from [JS](https://github.com/uclaacm/TeachLAJSBackend) to [Go](https://github.com/uclaacm/teach-la-go-backend). Here's a breakdown of the changes:

* Easier to follow calls to the backend in our modals and `fetch.js`.
    - Most of the endpoints use query parameters.
* Use HTTP response codes for validating responses as opposed to the `ok` field.
* Addition of `mostRecentProgramID` prop to the Main component.
    - Stores the document ID of the most recent program of an user as a prop. It is used for 
* Update `updateProgram` behavior
    - Update programs one by one instead of via array. We are only ever updating a single program at a time, so having it passed as an array didn't make much sense.
* Change our `deleteProgram` call to use both user ID and program ID (uid, pid) for program deletion instead of the old uid, docID, and name system.
    - This led to the creation of the `pid` prop in `ConfirmDeleteModal`, which is used when the modal makes the call to the backend.
* Tweak our `updateUserData` calls *slightly* (everything is sent as an object).
* Pass down a `sketchCount` prop to our `CreateSketchModal` used when creating a new sketch.

This PR is presently blocked on account of [this PR](https://github.com/uclaacm/teach-la-go-backend/pull/8) which updates the Go backend to work with the frontend changes made here. I'll unblock it on merge of said PR.

Stay healthy everyone and have a great spring break!


src/components/Sketches/index.js
```diff
@@ -169,13 +169,19 @@ class Sketches extends React.Component {
       onClose={() => this.setConfirmDeleteModalOpen(false)}
       sketchName={this.state.selectedSketch}
       sketchKey={this.state.selectedKey}
+      pid={
+        Array.from(this.props.programs)[this.state.selectedKey]
+          ? Array.from(this.props.programs)[this.state.selectedKey].uid
+          : ""
+      }
```
krashanoff: I'm not a huge fan of how I made the changes to this file, but these felt the most natural without modifying much else. Thoughts?

malsf21:
I'll give a more formal review after I finish my final, but just some quick notes:
* merging this & the `view-only` branch is going to be a bit of a fiasco, especially on the testing end
* we need to change the server addresses and/or `npm` actions to account for the new Go heroku URL
* looks like you've slightly changed the props structure for some components, so I'll have to investigate

I think it would probably be wise to incrementally implement these changes, i.e. allow `dev` to use the Go features but maintain "backwards compatibility" for production until we're 100% sure of the changes. Again, will discuss further in a bit.

krashanoff:
> merging this & the `view-only` branch is going to be a bit of a fiasco, especially on the testing end

I'm not opposed to simply basing my changes directly off of the `view-only` branch if that makes things easier on us. Between this PR and the view-only one, it's going to be an interesting time merging these all.

> we need to change the server addresses and/or npm actions to account for the new Go heroku URL

Will do.

> looks like you've slightly changed the props structure for some components, so I'll have to investigate

Here's the props I have added with brief descriptions:
* `pid` prop to ConfirmDeleteModal, which is the document ID of the program being deleted (used in the request to the backend).
* `sketchCount` prop to CreateSketchModal, the number of sketches currently listed (used when creating a new sketch and assigning an user's mostRecentProgram after creation).
* `mostRecentProgramID` to MainContainer, the document ID of the most recent program (used for saving the current program).

krashanoff:
Superseded by #253.

PR #245, opened 2020-03-25
View-Only: Fixes Redux State Corruption

malsf21:
PR Fix implemented by @jamieliu386:

The bug:
1. Open a sketch in view-only mode.
2. Go to the sketches page and try to download your most recently viewed sketch.
3. The sketch title and extension are as expected, but the code is the code from the view-only sketch.

This happens because the Redux state is corrupted with `mostRecentProgram`; this is a temporary bandaid that properly deals with the Redux state, but doesn't solve the core problem.

This PR fixes the bug by saving the original code for the most recently viewed program when View-Only is entered, and then restores it before the component unmounts.

@jamieliu386 and/or @krashanoff, I'd like to see if there's a quicker bandaid style fix for this - refactoring `mostRecentProgram` and our Redux user manager (i.e. clearing out the state when the user logs out, storing the right things in cookies, etc.) is likely out of scope for the View Only PR, but I'm concerned about the dependence on the lifecycle hook.

### DIFF
--- a/src/components/ViewOnly.js
+++ b/src/components/ViewOnly.js
@@ -31,11 +31,25 @@ class ViewOnly extends React.Component {
       code: "",
       loaded: false,
       notfound: false,
+      originalCode: "",
     };
   }
 
   componentDidMount = async () => {
+    const { ok: okOriginal, sketch: original } = await fetch.getSketch(
+      this.props.mostRecentProgram,
+    );
     const { ok, sketch } = await fetch.getSketch(this.props.programid);
+
+    if (!okOriginal) {
+      this.setState({ notfound: true });
+      return;
+    }
+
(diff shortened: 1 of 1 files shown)


malsf21:
Just added a check to make sure the user is logged in when we do the saving; if not, then it didn't work.

malsf21:
> Just got around to testing this out. The quick fix works. Will look into a more "bandaid style fix" for the root of the problem and report back.

Awesome, sounds good. If we can't find one super quickly, I don't mind merging this, then merging view only, then the go backend, and then reworking how mostRecentProgram and some of our Redux actions/state holders work.

malsf21:
Sounds good, thanks for the comment Leo. merging!

PR #253, opened 2020-04-08
Changes request patterns and response handling to match Go backend.

krashanoff:
Updates the frontend's backend request pattern to match our Go backend's.
* Favors HTTP response status instead of the "ok" field (eliminates use of `ok` field in backend responses).
* Uses HTTP methods for idiomatic requests (e.g., use `GET` for **get**ting your user data, and `PUT` for **put**ting new user data in the database)
* Adds the `userCreate` function to `fetch.js` (see [here](https://github.com/uclaacm/TeachLAFrontend/blob/go-requests/src/lib/fetch.js#L85)).
* Inherits a very strange antiquity from the JS backend (see below).

## getUserData

This is the weird thing. The JS backend, believe it or not, never had a `createUser` endpoint. Instead, it just made a call to `getUserData` and then the `getUserData` endpoint would create the user document if it didn't exist??? You can check out [the source code](https://github.com/uclaacm/TeachLAJSBackend/blob/master/lib/getUserData.js#L34) for yourself. This is real.

There's a lot to unpack there, so let's dive in.

### Why?

Well, if you take a peek at our [user creation flow](https://github.com/uclaacm/TeachLAFrontend/blob/go-requests/src/components/Login/CreateUserForm.js), specifically on [line 111](https://github.com/uclaacm/TeachLAFrontend/blob/go-requests/src/components/Login/CreateUserForm.js#L111), you'll see that we make a direct call to Firebase to create the user.

Is this wise? Kind of. It lets us get away with user creation without having to deal with hashing and storing it ourselves. The downside, though, is that we now have a promise-chained form submission that touches the database directly. I tried splicing a direct call to the [`user/create` endpoint](https://documenter.getpostman.com/view/10224331/SzYgSFSU?version=latest#f771bba1-b57c-44eb-9aa9-b45b39f96aac) on the new backend in there (say, in the [`then` after the call to firebase](https://github.com/uclaacm/TeachLAFrontend/blob/go-requests/src/components/Login/CreateUserForm.js#L114)), but to no avail.

### (less-than-ideal) Solution, Risks & Trade-offs

So how do we solve this problem? The fix I used was a rather dicey maneuver: replicate the JS backend's behavior, but on the frontend. That is, if someone has a cookie for an account that doesn't exist, and makes a call to [`getUserData`](https://github.com/uclaacm/TeachLAFrontend/blob/go-requests/src/lib/fetch.js#L17), then on failure the function will try to make a call to [`createUser`](https://github.com/uclaacm/TeachLAFrontend/blob/go-requests/src/lib/fetch.js#L85) before recursively calling itself. This is """"technically"""" safe, since `createUser` will fail if the backend is down, and succeed if the user already exists, but we can hypothetically get into an infinite recursion situation in the perfect storm.

I personally don't like that, and would like to eliminate this artifact, but I think that at this point we've been sitting on this PR for 6 months. It's "view-only" (#219) round two. It's time for it to get merged and handle the issue in another one.

***

This PR supersedes #235, and will hopefully - in time - **archive** [uclaacm/TeachLAJSBackend](https://github.com/uclaacm/TeachLAJSBackend). It will warrant a major semver bump, and deprecation of the [TeachLAJSBackend](https://github.com/uclaacm/TeachLAJSBackend). Prior to merge, the production Go backend will be deployed on Heroku.

### DIFF
--- a/src/components/Sketches/components/CreateSketchModal.js
+++ b/src/components/Sketches/components/CreateSketchModal.js
@@ -1,6 +1,6 @@
 import React from "react";
 import DropdownButton from "./DropdownButton";
-import ImageSelector from "../../common/ImageSelector"
+import ImageSelector from "../../common/ImageSelector";
 import {
   SketchThumbnailArray,
   LanguageDropdownValues,
@@ -46,7 +46,7 @@ class CreateSketchModal extends React.Component {
     });
   };
 
-  setNext = val => {
+  setNext = (val) => {
     this.setState({
       next: val,
       error: "",
@@ -111,7 +111,7 @@ class CreateSketchModal extends React.Component {
     return false;
   };
 
-  onFirstSubmit = e => {
+  onFirstSubmit = (e) => {
     e.preventDefault();
     if (this.badNameInput() || this.badLanguageInput()) {
       return;
@@ -119,7 +119,7 @@ class CreateSketchModal extends React.Component {
     this.setNext(true);
   };
 
-  onSecondSubmit = async e => {
+  onSecondSubmit = async (e) => {
     e.preventDefault();
 
     if (this.badThumbnailInput()) return;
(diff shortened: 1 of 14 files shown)


krashanoff:
Lord forgive me, work on this branch was getting so stale that I just went and redid it for a third time. It actually works now though. Unblocked.

src/lib/fetch.js
```diff
@@ -16,20 +16,26 @@ import constants from "../constants";
...
     mode: "cors", // no-cors, cors, *same-origin
   };
 
   try {
-    let result = await fetch(getUserDataEndpoint(uid, includePrograms), options);
-    let { ok, data, error } = await result.json();
-
+    const result = await fetch(getUserDataEndpoint(uid, includePrograms), options);
+    let ok = await result.ok;
+    if (!ok) {
```
willob99, line 29: Is it possible to check for the "user does not exist" error specifically before trying to create a new user?

src/components/Sketches/components/CreateSketchModal.js
```diff
@@ -135,23 +135,15 @@ class CreateSketchModal extends React.Component {
     try {
       fetch
         .createSketch(data)
-        .then(res => {
-          return res.json();
-        })
-        .then(json => {
-          if (!json.ok) {
```
willob99, line 142: Why did you stop checking .ok here and in TextEditor.js?

src/lib/fetch.js
krashanoff: Yeah we can check it via the response status (404 => user doesn't exist). Nice idea.

willob99 (approved):
Looks good to me!

PR #278, opened 2020-05-18
Add fork button

jamieliu386:
Adds the forking feature to a view-only sketch if the user is logged in.

fork button

After pressing the fork button, a modal appears asking the user to confirm the fork. 
confirmation

forking

Once the sketch is successfully forked, the user is prompted to visit their sketches page or close the modal.

sketch forked

If opened again, the modal remains at the final prompt.


jamieliu386:
I'm unsure about the styling of the fork button (what color?) and if the wording in the modal should be changed.

krashanoff:
I personally like the look of it as is right now, actually, but that's mostly because it's reminiscent of the GitHub fork look. Something that is a bit of a concern is that I feel like the icon is pretty unfamiliar for new coders and would warrant an explicit label, but then it wouldn't look as clean :sigh:

image

jamieliu386:
> Something that is a bit of a concern is that I feel like the icon is pretty unfamiliar for new coders and would warrant an explicit label

Yeah, I agree.

> but then it wouldn't look as clean :sigh:

A nice thing is that having an explicit label would look consistent with the Save button in the editor at least.

krashanoff:
> having an explicit label would look consistent with the Save button in the editor.

Yeah that's actually a huge plus. I'm leaning more and more towards giving it a label and leaving it at that. In terms of the manual/automatic redirect business, we still have to figure out a cleaner way of doing that -- might be a separate issue entirely since it touches on the editor and also the deleteModal.

malsf21:
So after further discussion, I think what's on live is from a feature-perspective reasonable enough. Definitely not in the right state of mind to do a code review, but @krashanoff take a look when you can (or tag someone in).

krashanoff:
I'll take code review on this one! I agree that things are looking good :+1:

Should have a review out by the end of the week.

src/components/TextEditor/components/TextEditor.js
```diff
@@ -34,6 +36,10 @@ class TextEditor extends React.Component {
       codeMirrorInstance: null,
       currentLine: 0,
       sketch: null,
+      showModal: false,
```
malsf21, line 39: Quick comment: this should probably be `showForkModal` just because we might/will add more modals later!

src/components/TextEditor/components/TextEditor.js
```diff
@@ -98,6 +112,87 @@ class TextEditor extends React.Component {
     this.setState({ currentLine: line });
   };
 
+  renderImageModal = () => {
```
malsf21, line 115: I assume this should be `renderForkModal`?

krashanoff: Still haven't gotten around to reviewing this, but from what I saw when I was looking over it earlier this week, I think it's just a typo since the fork button is based off of our deleteSketchModal.

jamieliu386:
@malsf21 Thanks for the review! I fixed the typos.

jamieliu386:
> I can't replicate this on the netlify deploy, but can consistently perform it locally. Anyone else see this issue?

I just tried out the sequence of steps (using one of my own sketches as well) and weirdly I was unable to reproduce the bug. Tagging @malsf21 to see if he can?

jamieliu386:
Ok, great! I resolved the merge conflicts and made sure the "share" button is hidden in view-only mode, so this feature should be good to go now!

krashanoff:
Phew! Glad we caught that one. I'll go ahead and merge this now that things are squared away. I'll be marking this as a minor release this week to commemorate the addition of the share and fork buttons. Thanks again Jamie! :+1:

PR #279, opened 2020-05-19
Add Share Sketch Button

malsf21:
This PR adds a "Share Sketch" button in all `TextEditor`, which opens a modal (a new component called `ShareSketchModal`) that allows the user to view the view-only URL to the current program, as well as copy-to-clipboard with a convenient button. I also added a library file `sketch.js`, which has a function called `constructShareableSketchURL` that generates the proper URL based on the Sketch ID and environment parameters. Note that this feature works whether or not the user is logged in, and/or viewing their own sketch.

Changes:
* created new component, `ShareSketchModal` in `src/components/TextEditor/components`
* created new library file, `sketch.js` in `src/lib/`, with function `constructShareableSketchURL`
* wrote tests for `constructShareableSketchURL`
* created a "standard" `.scss` file for Modal styles, in `Modals.scss`
* added button to header of `TextEditor` to trigger the modal open

Things to note:
* To implement the copy-to-clipboard, I used the [`Clipboard.writeText()`](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText) library. Note that this has no IE support; I have not implemented a fall-back, but I do create an error message.
* I have not written any tests for `ShareSketchModal`. I think eventually we should do a modal standardization pass + a set of smoke tests we can use.
* I think the button header is pretty stacked already, and with fork coming soon it'll be more crowded. We should figure out some way to make it more compact (either a dropdown or smaller icons), but that's out of scope of this PR.

### DIFF
--- a/src/lib/sketch.test.js
+++ b/src/lib/sketch.test.js
@@ -1,0 +1,23 @@
+import { constructShareableSketchURL } from "./sketch";
+
+describe("constructShareableSketchURL", () => {
+  it("works with localhost", () => {
+    expect(constructShareableSketchURL("owo", "localhost")).toBe("http://localhost:8080/p/owo");
+  });
+  it("works with editor.uclaacm.com", () => {
+    expect(constructShareableSketchURL("owo", "editor.uclaacm.com")).toBe(
+      "https://editor.uclaacm.com/p/owo",
+    );
+  });
+  it("somewhat gracefully handles no parameters", () => {
+    expect(constructShareableSketchURL()).toBe("http://localhost:8080/p/undefined");
+  });
+  it("reads from window.location.hostname", () => {
+    delete global.window.location;
+    global.window = Object.create(window);
+    global.window.location = {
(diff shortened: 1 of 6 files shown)


malsf21:
Cool, I ended up modifying the `constructShareableSketchURL` function to take an optional param that defaults to `window.location.hostname`, which makes testing easier. Also added a somewhat sketchy mock that just deletes the existing object and re-instantiates it: documentation about mocking variables (opposed to methods) is a bit shaky. Then, wrote some tests for the function.

LMK if it needs any other changes!

src/lib/sketch.test.js
```diff
@@ -0,0 +1,23 @@
+import { constructShareableSketchURL } from "./sketch";
+
+describe("constructShareableSketchURL", () => {
+  it("works with localhost", () => {
+    expect(constructShareableSketchURL("owo", "localhost")).toBe("http://localhost:8080/p/owo");
```
krashanoff, line 5: owo

PR #289, opened 2020-05-30
Fixes deleting mostRecentProgram in ConfirmDeleteModal

malsf21:
This PR fixes the nasty bug that corrupts our internal Redux state when you delete what happens to be the `mostRecentProgram`. We do this by adding a check when we delete the sketch: if the sketch's ID matches `mostRecentProgram`, we then update it with `setMostRecentProgram` with either the first item in the new programs list.

As a minor impl. note, we don't have to worry about a situation with only one sketch, as the editor component itself handles situations with zero sketches (and doesn't access mostRecentProgram). So, this fix doesn't change the existing bevhaiour in that scenario.

If we add more areas where we delete sketches (i.e. in the classes feature), we should probably make some action that does something to the effect of `deleteSketchAndUpdate` that handles all of these things at once, which reduces our cognitive load!

As an aside, we should probably check that `mostRecentProgram` is valid whenever we access it, and/or refactor many of its uses (as we've discussed, the editor depends too much on it).

### DIFF
--- a/src/components/Sketches/components/ConfirmDeleteModal.js
+++ b/src/components/Sketches/components/ConfirmDeleteModal.js
@@ -19,10 +19,10 @@ class ConfirmDeleteModal extends React.Component {
     try {
       fetch
         .deleteSketch(data)
-        .then(res => {
+        .then((res) => {
           return res.json();
         })
-        .then(json => {
+        .then((json) => {
           if (!json.ok) {
             this.setState({
               spinner: false,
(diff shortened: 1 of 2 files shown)


jamieliu386:
Hey Matt, looks good overall! However, I was able to reproduce the same bug in the edge case where I only had two sketches and deleted whichever one was the most recent program.

malsf21:
> Hey Matt, looks good overall! However, I was able to reproduce the same bug in the edge case where I only had two sketches and deleted whichever one was the most recent program.

Oop that seems bad, I'll take a look!

malsf21:
Okay, think I fixed it, @jamieliu386 can you take a look again?

jamieliu386:
Hey, sorry for the late review. Looks good on my end, too!

PR #292, opened 2020-06-03
Smaller "User Card" in Profile Panel

malsf21:
Draft PR; goal is to make a smaller user card (i.e. the user image + name), as the current one probably takes up a bit too much space. Also includes some general profile panel refactoring.

This should tie directly into then changing the navigation in the profile panel to be more condensed (and therefore support more navigation options).

### DIFF
--- a/src/components/common/ProfilePanel.js
+++ b/src/components/common/ProfilePanel.js
@@ -25,7 +25,7 @@ import "styles/Panel.scss";
 import Switch from "./Switch";
 import Footer from "./Footer";
 
-import ImageSelector from "./ImageSelector"
+import ImageSelector from "./ImageSelector";
 /**--------Props--------
  * togglePanel: function to be called when the panel is collapsed or opened
  */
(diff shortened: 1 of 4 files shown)


malsf21:
Wow, sorry I do not remember marking this ready for review, it's definitely still a draft. Hm, I think the empty space is somewhat awkward, though there are other solutions as well (in particular, spacing things out with flex maybe?). I think we still do need to make the name card smaller for "shorter" viewports, since we found w/ some testing that three navigation buttons pushed the theme toggle below the footer.

krashanoff:
No worries! I should have noticed your description mentioned it was still a draft. Regarding use of flex, I think that would probably make the most sense. If three navigation buttons messes with the theme button then it sounds like keeping things short is the play. Thanks Matt!

mattxwang:
@lumisphere902 just going through all the hanging PRs I've left, let me know if you want me to close this, hand it off, or finish it up when I have the time!

mattxwang:
Just doing a bit of self-consciousness - @lumisphere902 do you want me to close this?

PR #297, opened 2020-06-06
Adjust Input Validation Policies

malsf21:
This is a PR that follows up #283 in applying policy fixes to our input validations. Here is an exhaustive list of the changes that I've proposed:

* only allows alphanumeric characters and `-_.` in the username, which is consistent with the username policies of many other websites. key thing to note here is this now invalidates `@` in a username, which previously throws an error (as it can't be used in the LHS of an email address)
* increases the max character cap for usernames to 32: we don't face any storage issues here, and 20 seems slightly too restrictive (looking at twitter handles and the like)
* removes character restrictions on passwords. I think this is one of the most significant changes. There is no reason as to why spaces, periods, or a variety of other special characters cannot be used in passwords: they hinder password managers, and password restrictions are shown to make users forget their passwords easily and/or write them down physically. Plus, there is no technical need for the restriction: we just pipe the input into a hash function, and we never display it or query with it.
* increases the max character cap for passwords to 128: the current number of 20 is much too restrictive, both in terms of human generated passwords (such as the classic `correcthorsebatterystaple`) to the prevalence of password managers. Again, no technical issues here, as we aren't storing the password itself.
* allows for `-_.` (in addition to current policy) for display names.

~I am especially curious as to what our thoughts are on the username restriction; totally down to follow in the footsteps of Twitter or GitHub and allow for `-`, `_`, `.`, etc.~

Also, this PR removes a duplicate set of tests for `CreateUserForm` that are already covered in `validate`.


krashanoff:
Hey Matt! Thanks for following up on your other PR. After merging this I'll mark another minor release since we will have changed the login experience in rather considerable scale at that point.

***

> I am especially curious as to what our thoughts are on the username restriction; totally down to follow in the footsteps of Twitter or GitHub and allow for -, _, ., etc.

I think that this could be beneficial. It wouldn't mess with anything on our end as far as I know, and it adds a little more variety that kids might have seen in other social media.

malsf21:
Updated the changes!

PR #383, opened 2020-11-13
Refactor languages

lumisphere902:
- Make it easier to add new languages
- Fix downloads downloading blank file


krashanoff (commented):
Hey Timothy, this is looking good at the moment with the ViewOnly fixes taken into consideration. Only thing I have to comment on in terms of code review is that it would be better if you rebased this onto `master` just to keep #253 clean.

After that, I'll do a little more testing in the morning before giving the OK to merge.

krashanoff:
Reiterating: now that #253 has been merged, you'll need to either resolve merge conflicts by accepting all your existing changes, or rebase on master. Thanks for your patience Timothy.

src/util/languages/languages.js
```diff
@@ -0,0 +1,99 @@
...
+    extension: "py",
+    render: (code, showConsole) => CreatePythonDoc(btoa(code), showConsole),
+  },
+  {
+    value: "javascript",
+    display: "JavaScript",
+    icon: faJs,
+    codemirror: "javascript",
+    extension: "js",
+  },
```
krashanoff, line 22: Only comment here is that JavaScript is actually not officially supported by the editor. We could easily add this, but it'd have to be in another series of PRs since the backend won't accept this.

krashanoff (changes requested):
Timothy, thanks for the patience. This looks great. I'm really glad that we're finally cleaning up some of the editor after all this time.

Only one change required, then you've got the OK to merge. See below for more.

krashanoff (approved):

PR #399, opened 2020-12-04
Make classes page

willob99:


krashanoff (commented):
Hey Will! Thank you so much for sticking with this feature and Teach LA for so long. I have a few nitpicks in the details of the implementation, but looking over the entire PR from a big-picture, features-oriented take, this is looking really good.

Here are the big things that need addressing prior to merge.

## Back Button From Within Class View

Right now, if you open a class to view information about it, there's no easy way to go back:

image

A back button next to the class name is a must.

## Sketches Are Lost to the Void

Right now, when a teacher makes a sketch within a class, it doesn't register, and instead is lost to the void. I'm pretty sure this is just a small tweak in `fetch.js`, but I will also take a look at the backend to see if there's anything wrong over there.

## Join User to Class Breaks Class

When someone else joins a class, that class becomes unusable for them and all other students. This is likely just an implementation issue on the backend's part as far as I know. The only other thing I can think of is a corruption in Redux state on the frontend. I will look and see what needs fixing.

## Distinction Between User Account Types

This is another request for the backend. In the interim while we refactor user account distinction. We'll have to create a distinction between user account types such that a specific set of user accounts can access the instructor view of their classes and create new ones.

***

Aside from the few things mentioned above, all other major features appear to be in working order. Users can create new classes, leave classes, etc. with no trouble. More importantly, the application state doesn't appear to be corrupted from any arbitrary set of actions.

I know that you've been on this ticket for a *long* time and that graduation is coming up, so if you'd like someone else on the frontend team to take over for you in your stead, just let me know.

Great work, Will! Again, thank you.

Issue #430, opened 2021-01-31
Status line component


krashanoff:
In preparation for our collaborative coding endeavors, we need some way of indicating extra information to the user about what is going on in their editor. How about we pull a page out of more traditional code editors' notebooks? I'm thinking we add a status line to the editor. It should have information such as the user's current row, column, language, etc. This ticket covers just the creation and basic implementation of the component, not the exact spec.

Alternatively, this information could be integrated into the navbar in each editor instance.

For information on how to implement this, you can check out the CodeMirror [programming API docs](https://codemirror.net/doc/manual.html#api).

krashanoff:
Closing for the foreseeable future while collaborative coding's development plans are in suspense.

PR #432, opened 2021-02-04
status bar

PrateekSane:
created basic status bar. Displays whether editing or viewing. Shows line and column number. adds option to choose to display bar or not

### DIFF
--- a/src/components/TextEditor/components/StatusBar.js
+++ b/src/components/TextEditor/components/StatusBar.js
@@ -1,0 +1,12 @@
+import React from "react";
+
+const StatusBar = (props) => {
+  return (
+    <div className="status-bar-container">
+      <p className="status-bar-info">
+        {props.editing ? "Editing" : "Viewing"} line: {props.line + 1} col: {props.col + 1}{" "}
+      </p>
+    </div>
+  );
+};
+export default StatusBar;
--- a/src/components/TextEditor/components/TextEditor.js
+++ b/src/components/TextEditor/components/TextEditor.js
@@ -5,6 +5,7 @@ import sketch from "../../../lib/";
 
 import EditorRadio from "./EditorRadio.js";
 import ShareSketchModal from "./ShareSketchModal";
+import StatusBar from "./StatusBar";
 
 import { Button } from "reactstrap";
 import OpenPanelButtonContainer from "../../common/containers/OpenPanelButtonContainer";
@@ -37,12 +38,14 @@ class TextEditor extends React.Component {
     this.state = {
       codeMirrorInstance: null,
       currentLine: 0,
+      currentColumn: 0,
       sketch: null,
       showForkModal: false,
       forking: false,
       forked: false,
       redirectToSketch: false,
       showShareModal: false,
+      showStatus: true, //In case we have situations where we dont want to show status bar
     };
   }
 
@@ -102,17 +105,16 @@ class TextEditor extends React.Component {
     }
     this.props.setProgramCode(this.props.mostRecentProgram, newCode);
   };
-
   setCurrentLine = (cm) => {
     const { codeMirrorInstance, currentLine } = this.state;
-    let { line } = cm.getCursor();
+    let { line, ch } = cm.getCursor();
     if (codeMirrorInstance) {
       //removeLineClass removes the back highlight style from the last selected line
       codeMirrorInstance.removeLineClass(currentLine, "wrap", "selected-line");
       //addLineClass adds the style to the newly selected line
       codeMirrorInstance.addLineClass(line, "wrap", "selected-line");
     }
-    this.setState({ currentLine: line });
+    this.setState({ currentLine: line, currentColumn: ch });
   };
 
   renderForkModal = () => {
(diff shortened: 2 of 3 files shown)


krashanoff (commented):
Prateek, thanks for getting to work on this! Component itself is real simple, but I like the idea -- it's a great start. Right now, my comments are:
* The user should be able to toggle the status line while editing. The status line should be off by default.
* The status line should be toggled somewhere. I'm thinking we add a button or remove it for the time being? It will be integrated into the collaborative editor down the line.
* To that end, we will probably want to develop this feature on its own branch, and make pull requests against that. I will discuss this a little more in our meeting tonight.

To elaborate on the logic of why I suggested using the CodeMirror programming API in #430, I think that it would keep our styles a little tighter, since the status line would then update with light/dark mode and generally be easier to maintain. I would like your input on this if you feel a certain way, but I am focused on keeping the editor maintainable where possible.

.eslintcache
```diff
@@ -0,0 +1 @@
...js",[],"D:\\Projects\\TeachLA\\TeachLAFrontend\\src\\components\\EditorAndOutput\\EditorAndOutput.js",[],"D:\\Projects\\TeachLA\\TeachLAFrontend\\src\\components\\Sketches\\index.js",[],"D:\\Projects\\TeachLA\\TeachLAFrontend\\src\\components\\Sketches\\containers\\EditSketchModalContainer.js",[],"D:\\Projects\\TeachLA\\TeachLAFrontend\\src\\components\\common\\OpenPanelButton.js",[],"D:\\Projects\\TeachLA\\TeachLAFrontend\\src\\components\\Sketches\\components\\DropdownButton.js",[],"D:\\Projects\\TeachLA\\TeachLAFrontend\\src\\lib\\sketch.js",[],"D:\\Projects\\TeachLA\\TeachLAFrontend\\src\\components\\TextEditor\\components\\StatusBar.js",[],"D:\\Projects\\TeachLA\\TeachLAFrontend\\src\\components\\TextEditor\\components\\TextEditor.js",[],"D:\\Projects\\TeachLA\\TeachLAFrontend\\src\\components\\common\\ProfilePanel.js",[],"D:\\Projects\\TeachLA\\TeachLAFrontend\\src\\reducers\\uiReducer.js",[],"D:\\Projects\\TeachLA\\TeachLAFrontend\\src\\actions\\uiActions.js",[],"D:\\Projects\\TeachLA\\TeachLAFrontend\\src\\components\\common\\containers\\ProfilePanelContainer.js",[],"D:\\Projects\\TeachLA\\TeachLAFrontend\\src\\components\\TextEditor\\containers\\TextEditorContainer.js",[]]
```
krashanoff, line 1: This file can be safely deleted. I will add it to our `.gitignore` a little later.

src/components/TextEditor/components/StatusBar.js
```diff
@@ -1,10 +1,15 @@
...
+  const edit = props.editing;
+  const textStyle = { color: "white" };
+  if (props.theme === "light") {
+    textStyle.color = "black";
+  }
   return (
     <div className="status-bar-container">
-      <p className="status-bar-info">
-        {props.editing ? "Editing" : "Viewing"} line: {props.line + 1} col: {props.col + 1}{" "}
+      <p className="status-bar-info" style={textStyle}>
```
krashanoff, line 11: I still think that we could integrate this with our built-in styles with something like this instead:
```
<p className="status-bar-info" className={props.theme === "light" ? "light-status" : "dark-status"}>
```

src/components/common/ProfilePanel.js
```diff
@@ -316,6 +316,19 @@ class ProfilePanel extends React.Component {
     </div>
   );
 
+  renderShowStatusBar = () => {
+    let onToggle = () => {
+      console.log(this.props);
```
krashanoff, line 321: We can remove extraneous logging.
```suggestion
```

src/components/common/ProfilePanel.js
```diff
@@ -316,6 +316,19 @@ class ProfilePanel extends React.Component {
...
+      this.props.setStatusBar(!this.props.statusBar);
+    };
+    return (
+      <div className="panel-button">
+        <button className="statusbar-button" onClick={() => onToggle()}>
+          <span> Status Bar</span>
+        </button>
+      </div>
+    );
+  };
```
krashanoff, line 331: For now, let's mask the button from the editor, then we can add a button for it when we create the collaborative editor. Apologies if that wasn't clear. As a side note, using the same classes as `themeSwitch`, etc. would benefit its uniformity:

image

krashanoff (changes requested):
Hey Prateek, sorry for the delay on getting this review out. I have a few small changes. Many are nitpicky. The only significant one is that I think we should mask the toggle button from the editor for the moment, as this feature won't be useful outside of the collaborative environment. So, for now, we will have the status line ready to use when we cross that bridge.

Let me know if anything in my review is unclear. Otherwise, I'll keep an eye out for the next round of changes.

krashanoff (approved):
LGTM :+1:

PR #442, opened 2021-02-14
Adds ESLint linter + action, fixes style

mattxwang:
Draft for now! Def could use some help configuring all of this!


malsf21:
In troubleshooting this - so much of our test coverage just doesn't work right now. Going to probably delete a few files since the tests are completely nonsensical - any thoughts @krashanoff?

krashanoff:
> I suggest we delete a few files since the tests are completely nonsensical - any thoughts @krashanoff?

Agreed. Our tests on the frontend are severely lacking, and those that do exist are not much more than a smoke test. I'd say removing tests that aren't super rich is a good call.

mattxwang:
@lumisphere902 like the other ones I've tagged you in, let me know what you want me to do with this PR!

mattxwang:
Closed in favour of #507.

PR #501, opened 2021-04-30
Converted both common/ImageSelector.js #492  and  common/Switch.js #490 to functional component.

CharlesElloit:
Hi, I have converted both ImagesSelector and Switch to functional component.

### DIFF
--- a/src/components/common/Switch.js
+++ b/src/components/common/Switch.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 
 import "styles/Switch.scss";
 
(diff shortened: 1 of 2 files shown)


src/components/common/Switch.js
```diff
@@ -9,41 +9,34 @@ import "styles/Switch.scss";
...
-  }
-
-  componentDidUpdate(prevProps) {
-    if (this.props.on !== prevProps.on) {
-      this.setState({
-        on: this.props.on,
-      });
+
+const Switch = (props) => {
+  const [state, setState] = useState({
```
lumisphere902, line 14: This works, but does _on_ need to be nested inside a state object?

lumisphere902 (changes requested):
Hi Charles, thanks for contributing. There's just a small question I had about the Switch component. Other than that, LGTM to merge!

src/components/common/Switch.js
CharlesElloit: Well, not at all but i intend to follow state and setState naming convention from class to functional component and it's very clear for someone who hasn't use hook, considering that we are actually using the same state variable name throughout the component. Plus there is no any side effect of using that.
**But, if required, i will change it**

lumisphere902: This is unlikely to matter for this particular component, but in general I think separating state is better than a catch-all state

lumisphere902 (approved):
LGTM

PR #504, opened 2021-05-05
Add test-coverage script

lumisphere902:
Adds a test-coverage npm script, which also outputs the test coverage in a nice table. Obviously, tests won't appear super fast but it's something to slowly work towards.


jonathanbcarlson (approved):
Looks good, I'm glad we can finally see where we need to add tests!

PR #508, opened 2021-05-07
Add test coverage badge

reginawang99:
# Summary
- add test coverage to CI
     - only on push or pull request
- add test coverage badge
- remove enforce test coverage from `package.json`

# Decision Reasons
### Reason for removing `coverageThreshold` from `package.json`
Previously, in `package.json` we asked for 90% test coverage
```
    "coverageThreshold": {
      "global": {
        "branches": 90,
        "functions": 90,
        "lines": 90,
        "statements": 90
      }
    },
```
However, we currently only have ~ 15% coverage, which is nowhere near 90%
```
Jest: "global" coverage threshold for statements (90%) not met: 14.66%
Jest: "global" coverage threshold for branches (90%) not met: 17.44%
Jest: "global" coverage threshold for lines (90%) not met: 15.37%
Jest: "global" coverage threshold for functions (90%) not met: 11.78%
```
The continuous threshold error in running coverage causes complacency about our test coverage, and, in addition, having to add in `continue-on-error: true` for the ` run:  npm run test-coverage` step renders the `coverageThreshold` part a little useless in the first place. 

It is much more likely for someone to remember to add in the `coverageThreshold` into `package.json` than to remember to remove `continue-on-error: true` from the CICD pipeline. 

Thus, we should add in the `coverageThreshold` when we are actually using it.

### Line Coverage is also checked on pull request
Currently, we don't have any threshold for coverage, and it is theoretically not checking anything when we `npm test` (from `Node.js CI` action) in addition to when we perform line coverage checks. However, we should still check on pull request because
- it is nice to have the option to see the test coverage
- in the future, when we add in a threshold, we don't need to play around with the CI/CD again
- no downsides

### Use of Coveralls Decision is explained in the PR for the backend
[see here for more details](https://github.com/uclaacm/teach-la-go-backend/pull/54)

# Testing
- See [here](https://coveralls.io/github/uclaacm/TeachLAFrontend) 
- Also, CI clearly works


.github/workflows/nodejs.yml
```diff
@@ -1,24 +1,18 @@
...
+      uses: actions/checkout@v2
     - name: Use Node.js ${{ matrix.node-version }}
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
     - run: npm install
     - run: npm run prod_build --if-present
-    - run: npm test
-      env:
-        CI: true
```
reginawang99, line 24: See [here](https://docs.github.com/en/actions/reference/environment-variables) for details

lumisphere902 (approved):
LGTM. Try to change the coverage check, but if it doesn’t work out this still functions

.github/workflows/tests.yml
```diff
@@ -0,0 +1,31 @@
+name: Run tests
+on: [push]
+
+env: 
+  SHOULD_RUN_COVERAGE: ${{ !(github.event.push.branches == 'master' || github.event.issue.pull_request) }}
```
krashanoff, line 5: You can avoid writing environment variables by clarifying the trigger (see [here](https://futurestud.io/tutorials/github-actions-run-on-pull-request)).

krashanoff (commented):
Nice work Regina! Thank you.

Just here for a second opinion on the updates. With latest changes LGTM. I did leave a quick note on environment variables, though.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  coverage:
```
reginawang99, line 8:
```suggestion
  run-tests: 
```

.github/workflows/tests.yml
reginawang99: taken offline!

reginawang99:
Changed to Coveralls fails if coverage drops by more than 10%
You can modify in Coveralls like so: image

krashanoff (approved):
I'd say this LGTM. Thanks for the changes Regina!

PR #512, opened 2021-05-08
Add sass guidelines to stylelint

reginawang99:
# Summary
- add sass guidelines to style lint (from https://github.com/bjankord/stylelint-config-sass-guidelines)
- add `yarn.lock` to `.gitignore`
- move `stylelint ignoreFile` to `.stylelintignore` since `ignoreFile` wasn't actually working (tested before and after)
- stricter stylelinting, and getting rid of duplicate rules

# For Reviewer
- only need to check out first 4 files! (~40 lines)

# Next
ignoring all violating rules for now. I would recommend bringing back these three rules:
- `"scss/at-import-partial-extension-blacklist": null,`
-  `"declaration-property-value-disallowed-list": null,`
- `"selector-class-pattern": null`

they should be pretty easy changes, and i can do this in the next commit!!


src/styles/Resizer.scss
```diff
@@ -2,41 +2,41 @@
```
reginawang99, line 2: i dont think any files are actually using the `Resizer` class?? it's in `EditorAndOutput.js` but no one ever specifies for the `Resizer` class?

reginawang99:
@mattxwang oop i attached the wrong link for `stylelint-config-sass-guidelines` but i removed it cuz it's already in the guidelines!!

https://github.com/bjankord/stylelint-config-sass-guidelines/blob/fe3ea58083851a68480c4615f9f9e4b0a5e66da8/index.js#L5

do you want me to add `at-rule-no-unknown` back in anyhow?

mattxwang:
> @mattxwang oop i attached the wrong link for `stylelint-config-sass-guidelines` but it's already in the guidelines!!
> 
> https://github.com/bjankord/stylelint-config-sass-guidelines/blob/fe3ea58083851a68480c4615f9f9e4b0a5e66da8/index.js#L5
> 
> do you want me to add `at-rule-no-unknown` back in anyhow?

Oh, I didn't catch this - 100% my bad then! Good on my end (though I'm not an official reviewer)

src/styles/global.scss
```diff
@@ -33,20 +33,20 @@ h6 {
 }
 
 .teachla-green {
+  background-clip: text;
+  background-clip: text;
```
krashanoff, line 37:
```suggestion
  background-clip: text;
```

.stylelintrc.json
```diff
@@ -1,11 +1,12 @@
...
   "rules": {
-    "at-rule-no-unknown": null,
-    "scss/at-rule-no-unknown": true
+    "color-named": null,
+    "selector-max-id": null,
+    "selector-no-qualifying-type": null,
+
+    "scss/at-import-partial-extension-blacklist": null,
+    "declaration-property-value-disallowed-list": null,
+    "selector-class-pattern": null
```
krashanoff, line 10: I do think we might want to preserve our `"ignoreFiles"` field sans Typescript.

I'm not too familiar with Stylelint's rule extensions. Do our rules extend the `stylelint-config-standard` implicitly?

src/styles/Login.scss
```diff
@@ -62,82 +62,83 @@
...
 
 .login-form-input {
-  font-size: 1.5em;
-  line-height: normal;
-  padding: 0.2rem 0.5rem;
-  border: 0;
-  border-radius: 8px;
+  appearance: none;
+  appearance: none;
+  appearance: none;
```
krashanoff, line 111:
```suggestion
  appearance: none;
```

src/styles/Resizer.scss
```diff
@@ -2,41 +2,41 @@
 
 .Resizer {
   background: #000;
+  background-clip: padding;
+  background-clip: padding;
+  background-clip: padding-box;
+  box-sizing: border-box;
+  box-sizing: border-box;
+  box-sizing: border-box;
```
krashanoff, line 10:
```suggestion
  background-clip: padding-box;
  box-sizing: border-box;
```

I'm actually a little worried about this change since `background-clip` actually [takes different values](https://developer.mozilla.org/en-US/docs/Web/CSS/background-clip) than the per-browser webkit properties.

src/styles/Resizer.scss
```diff
@@ -2,41 +2,41 @@
...
-  box-sizing: border-box;
-  -moz-background-clip: padding;
-  -webkit-background-clip: padding;
-  background-clip: padding-box;
 }
 
 .Resizer:hover {
-  -webkit-transition: all 2s ease;
+  transition: all 2s ease;
   transition: all 2s ease;
```
krashanoff, line 17:
```suggestion
  transition: all 2s ease;
```

krashanoff (commented):
Hey Regina, thanks for handling devops on the frontend. Everything looked fine and I'll likely leave the final word to @lumisphere902, but I did point out a few odd things down in my thread, along with some minor changes to remove duplicates and the like.

.stylelintrc.json
reginawang99: yup! i created a `.stylelintignore` file since the `ignoreFiles` thing here wasn't actually working. 
- I believe stylelint is a lil finnicky about where `ignoreFiles` goes or something, so i didnt want to deal with it, and instead created the ignore file

src/styles/Login.scss
reginawang99: huh im kinda sus of all these changes, ill take a look at whats happening

.stylelintrc.json
krashanoff: Sounds good then. Thanks for doing that.

.stylelintrc.json
```diff
@@ -1,11 +1,12 @@
 {
-  "extends": "stylelint-config-standard",
-  "plugins": [
-    "stylelint-scss"
-  ],
-  "ignoreFiles": ["**/*.ts", "**/*.tsx", "dist/*", "node_modules/*", "coverage/*"],
+  "extends": "stylelint-config-sass-guidelines",
```
reginawang99, line 2: @krashanoff nice catch about not extending `"stylelint-config-standard"`, ill add it to the extends array!

src/styles/Resizer.scss
reginawang99: hm ill disable the rule then

reginawang99: i disabled the rule that caused this and reverted this to what it was originally!

reginawang99:
decided to extend off of `stylelint-config-recommended` and `stylelint-config-sass-guidelines` rather than `stylelint-config-standard` because the standard one has conflicts with the sass one. The `recommended` one is pretty base case and covers most  essential bases

krashanoff (approved):
> decided to extend off of `stylelint-config-recommended` and `stylelint-config-sass-guidelines` rather than `stylelint-config-standard`

Sounds fine to me. Looking over the changes Stylelint made in its passthrough, I see no issues.

For the finnickier rules, we can just manually go through them. `box-sizing`, for example, is [widely supported](https://caniuse.com/css3-boxsizing). This is outside of the scope of this PR, though, and I believe we are fine to merge.

PR #513, opened 2021-05-09
add husky lint hook and stricter linting

### LINKED ISSUE #446: Reconfigure Husky/Lint-staged with Stylelint/ESLint
After #442 and #439, it might be worth re-configuring husky and lint-staged (which I don't think 100% work properly right now anyways) to run a precommit hook to autofix any style problems and prevent the user from committing egregious code format/lint errors! AFAIK with the new version of husky/lint-staged you may need to uninstall and reinstall the package to run some of the post-install scripts.

reginawang99:
closes #446


.env
```diff
@@ -1 +1,2 @@
 PORT=8080
+DISABLE_ESLINT_PLUGIN=true
```
lumisphere902, line 2: I know it's probably getting annoying but newline please :(

lumisphere902 (approved):
LGTM

.env
reginawang99: oh my bad, that was my fault

PR #525, opened 2021-05-20
fix SketchBox language

lumisphere902:
Fixes a couple issues with sketches when interacting from the main Sketches page. (Icons and download not working)


krashanoff (approved):
LGTM, nice catch.

PR #531, opened 2021-05-24
drop down button fix

### LINKED ISSUE #486: Consolidate DropdownButton components
Currently, we have two DropdownButton components: [common/DropdownButton.js](https://github.com/uclaacm/TeachLAFrontend/blob/master/src/components/common/DropdownButton.js) and [Sketches/components/DropdownButton.js](https://github.com/uclaacm/TeachLAFrontend/blob/master/src/components/Sketches/components/DropdownButton.js). Realistically, only one of these needs to exist. In the process, we should change it into a single functional component w/ hooks

ldalton02:
closes #486
Combines two separate dropdownbutton components into a singular component and converts to a functional component


krashanoff (commented):
Nice call on this refactor, Luke! Thanks for getting this done.

This looks good to me, but I'll loop in @lumisphere902 for a second opinion before it's okay to merge. One more thing: if you could update the title/description on this PR it would help out in analyzing the commit history for the frontend.

lumisphere902 (approved):
LGTM

src/components/common/DropdownButton.js
```diff
@@ -12,66 +10,48 @@ import {
...
 
-    if (this.props.onSelect && result) {
-      this.props.onSelect(program);
+    if (onSelect && result) {
+      onSelect(program);
     }
   };
 
-  renderDropdownItems = () =>
+  const renderDropdownItems = () =>
```
reginawang99, line 34: theres a different `renderDropdownItems` for the Sketches button!

src/components/common/DropdownButton.js
```diff
@@ -12,66 +10,48 @@ import {
...
-}
+    ));
+  return (
+    <div className="editor-language-dropdown">
+      <Dropdown isOpen={dropdownOpen} toggle={() => toggleHandler(dropdownOpen)}>
+        {/* HACK: disables the colors entirely, makes the dropdown transparent */}
+        <DropdownToggle className="btn-language-dropdown" color="" caret>
+          <div className="editor-language-dropdown-closed-content">
+            <FontAwesomeIcon icon={currentLanguage.icon} fixedWidth /> {displayValue}
+          </div>
```
reginawang99, line 49: check this! its a little different from the Sketches button

reginawang99 (dismissed):
looks good to me so far! 

i think you mightve missed some differences between the buttons. id recommend putting the two buttons through a diff checker (like [this](https://www.diffchecker.com/diff)) and noticing the differences!! (you might need to pass in an additional prop to denote whether its a common or sketches button!)

and +1 with leo
> One more thing: if you could update the title/description on this PR it would help out in analyzing the commit history for the frontend.

reginawang99:
@krashanoff / @lumisphere902 

Looking at this, I don't really see why we are combining these buttons. One way I can think of to combine these 2 seamlessly is passing in:
- Optional `className` string
- Optional child `dropdownToggle` JSX element (which overrides default)
- Optional `renderDropdownItems` method (which overrides default)

And also creating a reusable `SketchesDropdownButton` that fills in the 3 optional things. (To be reused across components in the `Sketches` folder)

But that's basically all the logic and just serves to make this component more complex. Or we can just leave like this 😅 

Any thoughts?

lumisphere902:
I still think that having multiple DropdownButton components is a bad idea. The goal would not be to have an isSketch prop passed in, but rather to have stuff like className, etc passed in. I don't think dropdownToggle or renderDropdownItems methods are too far removed from each other that they need to be completely separate; I think it should be possible to have an optional icon, and a few other changes to consolidate the two versions.

reginawang99:
I agree that I don't prefer having an `isSketch` prop, but `renderDropdownItems` might be a necessary pass in?
     + also, there are several components that require a `Sketches` button (which need to be the same)

Do you want to take this review? I feel like you might have stronger opinions that I do, and I don't want to have luke change things between my and your review

krashanoff:
The premise of this refactor is good. I don't think we should just abandon it; having name ambiguities is bad enough in and of itself. For the actual implementation details, I trust Timothy. I will suggest, though, that one could use `props.children` instead of `dropdownItems`. Trade-off here is that we're introducing an opaque level of abstraction. Passage of `className` is no issue in my opinion, since it dictates the styling of the parent container.

lumisphere902:
For renderDropDownItems, I think the best way to do it is to create a more generic "data" object with values like display, value, icon. Then, for the non-sketch dropdown, you can convert the program data to that format.  (e.g. value is program.name). I think it's best to have a generic common component, with an interface that the sketches/non-sketches conforms to. If there's a lot of repeated code, you can make a util function to convert it.

src/components/common/containers/DropdownButtonContainer.js
```diff
@@ -1,34 +1,65 @@
 import { connect } from 'react-redux';
```
lumisphere902, line 1: Considering that we use `DropdownButton` without this container, maybe we could move it to components/TextEditor/containers. Most of the logic in here is pretty specific to the TextEditor anyway.

src/components/common/DropdownButton.js
```diff
@@ -1,77 +1,82 @@
...
-              {this.props.displayValue}
-            </div>
-          </DropdownToggle>
-          <DropdownMenu>{this.renderDropdownItems()}</DropdownMenu>
-        </Dropdown>
-      </div>
-    );
-  }
-}
+/*
```
lumisphere902, line 58: Obviously these should eventually be removed, but I understand having it here for reference.

lumisphere902 (changes requested):
Good progress, but there are a few ways I think that the component could be improved. PTAL and let me know if you have any questions! :)

src/components/common/DropdownButton.js
```diff
@@ -1,77 +1,82 @@
...
-  render() {
-    // let value = this.props.displayValue
-    // if(this.props.dirty){
-    //   value = (<span>&#8226;{this.props.displayValue}</span>)
-    // }
+  return (
+    <div className={parentDisplay}>
+      <Dropdown isOpen={dropdownOpen} toggle={() => toggleHandler(dropdownOpen)}>
+        {/* HACK: disables the colors entirely, makes the dropdown transparent*/}
+        <DropdownToggle className={toggleClass} color={toggleColor} size={toggleSize} caret>
```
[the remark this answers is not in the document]
ldalton02: Hey, I was having some trouble making this happen and I think I might be messing up the syntax a bit. In the DropdownButton container, I simply created an object called toggleprops that is as follows:

const toggleProps = {
    class: 'btn-language-dropdown',
    color: '',
    size: '',
  };

This works fine when passed in from the container. However, in the sketch classes, when I try to pass in an identical object, the data for some reason does not transfer over correctly. Could you let me know what I'm missing?

This is what I put as a prop in the sketch classes:
toggleProps={{ class: '', color: 'blue', size: 'lg'}}

src/components/common/DropdownButton.js
```diff
@@ -1,77 +1,82 @@
...
+    setdropdownOpen(!prevVal);
   };
 
-  selectLanguage = (program) => {
-    let result = true;
-    if (this.props.dirty) {
-      result = window.confirm('Are you sure you want to change programs? You have unsaved changes');
-    }
+  const renderDropdownItems = () =>
+    children.map(({ display, value }) => {
```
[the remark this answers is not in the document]
ldalton02: Just to clarify, for "children" it would be alright if I processed it outside and children was already just an array containing all the icons and display names etc?

src/components/common/DropdownButton.js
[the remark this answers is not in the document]
reginawang99: this is how you pass props & spread!
``` 
const toggelProps = {className: "btn-language-dropdown", something: "value", somethingElse: "value2"};
```
and to spread, you can
```
<DropdownToggle {...toggelProps} caret>
```

src/components/common/DropdownButton.js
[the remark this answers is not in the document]
reginawang99: in general, `children` refers to the components inside of the component which `children` or `child` is passed to.
so children would be something more used like
```
<DropdownButton>
   <child>asdfasdf</child>
</DropdownButton>
```
as per what timothy said in a different comment!

so with that, if it was just pure JSX, then yes, calling it "children" would be accurate

ldalton02:
quick comment on the most recent pull request: I misinterpreted what you said about the children prop at first and changed it so that all processing of data is done in the parent containers instead of within the dropdownbutton class. Do you like what I did with the processing or think it's bad practice? Other than that, I believe I addressed all of the other changes requested

src/components/common/DropdownButton.js
```diff
@@ -3,80 +3,45 @@ import React, { useState } from 'react';
...
-    onSelect,
-    children,
-    defaultOpen,
-    displayValue,
-    displayClass,
-    toggleClass,
-    toggleColor,
-    toggleSize,
-  } = props;
+  const { children, icon, defaultOpen, displayValue, displayClass, ...toggleProps } = props;
```
lumisphere902, line 17: Instead of taking the rest of the props as the toggleProps, I think that it should be explicitly passed in as an object toggleprops, and then spread from that.

src/components/Sketches/components/CreateSketchModal.js
```diff
@@ -43,6 +46,18 @@ class CreateSketchModal extends React.Component {
     });
   };
 
+  changeLanguage = (lang) => {
+    this.setState({ language: lang });
+  };
+
+  dropDownValues = LanguageDropdownValues.map(({ display, value }) => {
```
lumisphere902, line 53: Sorry for being unclear earlier. Don’t like making you undo work, but think about it this way. In the future if you or someone else wants to include a dropdown button in a page, it wouldn’t make sense for them to essentially write the rendering logic again.

src/components/Sketches/components/CreateSketchModal.js
```diff
@@ -247,14 +262,13 @@ class CreateSketchModal extends React.Component {
...
-              children={LanguageDropdownValues}
-              onSelect={(lang) => this.setState({ language: lang })}
               displayValue={this.state.language.display || LanguageDropdownDefault.display}
               displayClass={'sketches'}
-              toggleClass={''}
-              toggleColor={'primary'}
-              toggleSize={'lg'}
-            />
+              icon={null}
+              {...this.toggleProps}
```
lumisphere902, line 268: This should be passed in like toggleProps={this.toggleProps}

lumisphere902 (changes requested):
Yeah, sorry about the misunderstanding. I do think it’s worth changing the rendering logic back to inside the common component though.

ldalton02:
No problem, my mistake again for misreading what you put. I renamed it accordingly to reflect the items not being "children" of the dropdown and rendered inside the dropdown again. Fixed the toggleprops spreading as well

reginawang99:
@lumisphere902 just a bop in case you wanted to do this!

src/components/common/DropdownButton.js
```diff
@@ -1,77 +1,76 @@
...
-      <div className="editor-language-dropdown">
-        <Dropdown
-          isOpen={this.state.dropdownOpen}
-          toggle={() => this.toggleHandler(this.state.dropdownOpen)}
+  return (
+    <div className={dropDownParentClass}>
+      <Dropdown isOpen={dropdownOpen} toggle={() => toggleHandler(dropdownOpen)}>
+        {/* HACK: disables the colors entirely, makes the dropdown transparent*/}
+        <DropdownToggle
+          className={togglePropsValues.class}
```
lumisphere902, line 61: You can directly spread toggleProps here

src/components/common/DropdownButton.js
```diff
@@ -1,77 +1,76 @@
...
-    // let value = this.props.displayValue
-    // if(this.props.dirty){
-    //   value = (<span>&#8226;{this.props.displayValue}</span>)
-    // }
+  const renderDropdownItems = () =>
+    DropdownItems.map(({ display, value, icon }) => {
+      return (
+        <DropdownItem key={value} onClick={() => onSelect({ display, value, dirty })}>
+          <FontAwesomeIcon icon={icon} fixedWidth />
+          <span style={{ marginLeft: '10px' }}> {display} </span>
```
lumisphere902, line 51: Does the styling look correct for both w/ icon and w/o icon?

src/components/Sketches/components/EditSketchModal.js
```diff
@@ -192,13 +195,16 @@ class EditSketchModal extends React.Component {
...
               <DropdownButton
-                dropdownItems={LanguageDropdownValues}
-                onSelect={(lang) => this.setState({ newLanguage: lang })}
                 displayValue={
                   this.state.newLanguage !== -1
                     ? this.state.newLanguage.display
                     : this.props.sketchLang
                 }
+                displayClass={'sketches'}
+                icon={null}
```
lumisphere902, line 204: You shouldn’t need to explicitly set icon to null

lumisphere902 (changes requested):
Hey Luke, sorry about the delay on the review. It mostly looks good; just a few smaller things left to change here. Make sure to test both places that the DropdownButton is used, and compare it with previous behavior. With that done, we should be good. :)

ldalton02:
I switches the margin to be applied to the icon so that the margin between the two elements will only appear in the dropdown within the editor.
I also just tested the behavior and the dropdown seems to look and function in both cases identically to the version live on the editor website

src/components/common/DropdownButton.js
```diff
@@ -1,77 +1,66 @@
...
-    return (
-      <div className="editor-language-dropdown">
-        <Dropdown
-          isOpen={this.state.dropdownOpen}
-          toggle={() => this.toggleHandler(this.state.dropdownOpen)}
+  return (
+    <div className={dropDownParentClass}>
+      <Dropdown isOpen={dropdownOpen} toggle={() => toggleHandler(dropdownOpen)}>
+        <DropdownToggle
+          className={toggleProps.class}
```
lumisphere902, line 51: Sorry if I was unclear last time. I meant that you could spread toggleProps like {...toggleProps}. This way, if someone wants to modify the props passed to DropdownToggle (e.g. add a new one) they do not need to change DropdownButton.

src/components/TextEditor/containers/DropdownButtonContainer.js
```diff
@@ -0,0 +1,56 @@
+import { connect } from 'react-redux';
+import { setMostRecentProgram } from '../../../actions/userDataActions.js';
+import { getLanguageData } from '../../../util/languages/languages.js';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
```
lumisphere902, line 4: I don't believe this import is used.

lumisphere902 (changes requested):
Hey, just a few more things I commented on.

src/components/common/DropdownButton.js
```diff
@@ -1,77 +1,66 @@
...
-    return (
-      <div className="editor-language-dropdown">
-        <Dropdown
-          isOpen={this.state.dropdownOpen}
-          toggle={() => this.toggleHandler(this.state.dropdownOpen)}
+  return (
+    <div className={dropDownParentClass}>
+      <Dropdown isOpen={dropdownOpen} toggle={() => toggleHandler(dropdownOpen)}>
+        <DropdownToggle
+          className={{ ...toggleProps }.class}
```
lumisphere902, line 51: What I'm trying to say is that you can replace all of these props (the whole className=, color=, size=) with {...toggleProps}. You would have to change class -> className however. Something like
`<DropdownToggle caret {...toggleProps} />`

ldalton02: Oh ok, that makes sense! My apologies, I didn't know that props could be spread like that. Sorry for the back and forth, hope this last push fixes it

lumisphere902 (approved):
LGTM. Thanks for all the work you did on this! And I apologize for being slow with getting back to you.

ldalton02:
No problem, glad to help out

PR #555, opened 2021-07-15
Refactored ProfilePanel.js

TrevorGuo:
Refactored ProfilePanel.js from class based to functional.

Notes:
   - Line 65 has an unused function "handleEditImageClick"
   - Line 155 error was not declared previously as a state variable. I added it in, and set its initial value to an empty string.

### DIFF
--- a/src/components/common/ProfilePanel.js
+++ b/src/components/common/ProfilePanel.js
@@ -192,7 +192,7 @@ function ProfilePanel(props) {
           onMouseLeave={() => setNameHover(false)}
           onClick={handleEditNameClick}
         >
-          <div className="panel-name-text">{displayName || 'Joe Bruin'}</div>
+          <div className="panel-name-text">{name || 'Joe Bruin'}</div>
           {nameIsHovering && (
             <button className="edit-icon-image" onClick={handleEditNameClick}>
               <FontAwesomeIcon icon={faEdit} />
--- a/src/components/common/ProfilePanel.js
+++ b/src/components/common/ProfilePanel.js
@@ -59,7 +59,7 @@ function ProfilePanel(props) {
   const [name, setName] = useState(displayName);
   const [selectedImage, setSelectedImage] = useState('');
   const [displayNameMessage, setDisplayNameMessage] = useState('');
-  const [error, setError] = useState('');
+  const [error, setError] = useState(''); // setError is never used
 
   //State Changing functions
   const handleOpenModal = () => {
--- a/src/components/common/ProfilePanel.js
+++ b/src/components/common/ProfilePanel.js
@@ -35,13 +35,28 @@ import Switch from './Switch';
  */
 
 function ProfilePanel(props) {
+  //Props
+  const {
+    photoName,
+    setPhotoName,
+    displayName,
+    setDisplayName,
+    contentType,
+    theme,
+    onThemeChange,
+    togglePanel,
+    left,
+    screenHeight,
+    uid,
+  } = props;
+
   //States
   const [nameIsHovering, setNameHover] = useState(false);
   const [imageIsHovering, setImageHover] = useState(false);
   const [editingName, setEditingName] = useState(false);
   const [showModal, setShowModal] = useState(false);
   const [nameSubmitted, setNameSubmitted] = useState(false);
-  const [name, setName] = useState(props.displayName);
+  const [name, setName] = useState(displayName);
   const [selectedImage, setSelectedImage] = useState('');
   const [displayNameMessage, setDisplayNameMessage] = useState('');
   const [error, setError] = useState('');
@@ -49,7 +64,7 @@ function ProfilePanel(props) {
   //State Changing functions
   const handleOpenModal = () => {
     setShowModal(true);
-    setSelectedImage(props.photoName);
+    setSelectedImage(photoName);
   };
 
   const handleCloseModal = () => {
@@ -61,11 +76,6 @@ function ProfilePanel(props) {
     setEditingName(true);
   };
 
-  //Function not used
-  const handleEditImageClick = () => {
-    setShowModal(true);
-  };
-
   const onNameChange = (e) => {
     setName(e.target.value);
   };
@@ -75,11 +85,11 @@ function ProfilePanel(props) {
 
     const { ok, message } = isValidDisplayName(name);
     if (!ok) {
-      setName(props.displayName);
+      setName(displayName);
       setEditingName(true);
       setDisplayNameMessage(message);
     } else {
-      props.setDisplayName(name);
+      setDisplayName(name);
       setEditingName(false);
       setNameSubmitted(true);
       setDisplayNameMessage('');
@@ -95,7 +105,7 @@ function ProfilePanel(props) {
    */
   const onImageSubmit = () => {
     // SEND IMAGE NAME TO BACKEND, CHANGE IMAGE
-    props.setPhotoName(selectedImage);
+    setPhotoName(selectedImage);
     handleCloseModal();
     setSelectedImage('');
   };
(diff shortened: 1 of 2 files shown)


src/components/common/ProfilePanel.js
```diff
@@ -26,58 +34,57 @@ import Switch from './Switch';
...
+  const [nameIsHovering, setNameHover] = useState(false);
+  const [imageIsHovering, setImageHover] = useState(false);
+  const [editingName, setEditingName] = useState(false);
+  const [showModal, setShowModal] = useState(false);
+  const [nameSubmitted, setNameSubmitted] = useState(false);
+  const [name, setName] = useState(props.displayName);
+  const [selectedImage, setSelectedImage] = useState('');
+  const [displayNameMessage, setDisplayNameMessage] = useState('');
+  const [error, setError] = useState('');
+
```
reginawang99, line 48: nit: let's destructure the props so that we know what props are actually used here. It also helps other know what props are required when using this component.

```
const {displayName, setDisplayName, ...} = props
```
and then we can use `displayName` instead of `props.displayName`

src/components/common/ProfilePanel.js
```diff
@@ -242,39 +246,41 @@ class ProfilePanel extends React.Component {
...
 
     return <div className="panel-buttons">{panelButtons}</div>;
   };
 
-  renderThemeSwitch = () => {
+  const renderThemeSwitch = () => {
     const onToggle = (on) => {
-      this.props.onThemeChange(on ? 'light' : 'dark');
+      props.onThemeChange(on ? 'light' : 'dark');
     };
```
reginawang99, line 272: nit: you can remove `{}` if you want since it's one line arrow function, but not important

src/components/common/ProfilePanel.js
```diff
@@ -228,7 +232,7 @@ class ProfilePanel extends React.Component {
     </Link>
   );
```
reginawang99, line 233: there's a lot of these buttons duplicated. it might be good to have a `renderProfilePanelButton` function and specify with a docstring which buttons can be rendered with this function. It should probably take in `icon`, and `name` or something like that

Feel free to not do this if you don't want to since it's out of scope, but it'd be nice to have

reginawang99 (changes requested):
looks mostly good to me!! We probably need to make the tests work, so you should probably look into that a bit.

Another thing if you want to look into it, I notice there's a ton of duplication which is bloating the file. If you want to do some clean code stuffs, you can clean that up + I can think of other ways to do so after your first pass!

Feel free to ask me questions for anything :)

src/components/common/ProfilePanel.js
```diff
@@ -26,58 +34,57 @@ import Switch from './Switch';
...
-    this.setState({ editingName: true });
+  const handleEditNameClick = () => {
+    setEditingName(true);
   };
 
-  handleEditImageClick = () => {
-    this.setState({ showModal: true });
+  //Function not used
+  const handleEditImageClick = () => {
+    setShowModal(true);
```
reginawang99, line 66: if we aren't using it, we can remove it

src/components/common/ProfilePanel.js
```diff
@@ -166,45 +183,42 @@ class ProfilePanel extends React.Component {
...
-          onMouseLeave={() => this.setState({ nameIsHovering: false })}
-          onClick={this.handleEditNameClick}
+          onMouseEnter={() => setNameHover(true)}
+          onMouseLeave={() => setNameHover(false)}
+          onClick={handleEditNameClick}
         >
-          <div className="panel-name-text">{this.props.displayName || 'Joe Bruin'}</div>
-          {this.state.nameIsHovering && (
-            <button className="edit-icon-image" onClick={this.handleEditNameClick}>
+          <div className="panel-name-text">{name || 'Joe Bruin'}</div>
```
reginawang3495, line 195: Is this change intentional?
```suggestion
          <div className="panel-name-text">{displayName || 'Joe Bruin'}</div>
```

src/components/common/ProfilePanel.test.js
```diff
@@ -102,40 +100,40 @@ describe('ProfilePanel', () => {
...
     component.find('.panel-name').simulate('mouseenter');
-    expect(component.state().nameIsHovering).toBe(true);
+    expect(component.find('.edit-icon-image').exists()).toBe(true);
 
     // click the pencil icon (changes to input)
-    expect(component.state().editingName).toBe(false);
+    expect(component.find('.panel-edit-container').exists()).toBe(false);
     component.find('.edit-icon-image').simulate('click');
-    expect(component.state().editingName).toBe(true);
+    // expect(component.state().editingName).toBe(true);
```
reginawang3495, line 110:
```suggestion
```

src/components/common/ProfilePanel.test.js
```diff
@@ -102,40 +100,40 @@ describe('ProfilePanel', () => {
...
     expect(component.find('.panel-edit-container')).toHaveLength(1);
 
     // check the input value starts as 'Mark', type in the input 'Not Mark',
     // check that the input value and state changes to 'Not Mark'
-    expect(component.state().name).toBe('Mark');
+    // expect(component.state().name).toBe('Mark');
```
reginawang3495, line 115:
```suggestion
```

src/components/common/ProfilePanel.test.js
```diff
@@ -102,40 +100,40 @@ describe('ProfilePanel', () => {
...
     expect(component.find('.panel-edit').props().value).toBe('Mark');
     component.find('.panel-edit').simulate('change', {
       target: {
         value: 'Not Mark',
       },
     });
     expect(component.find('.panel-edit').props().value).toBe('Not Mark');
-    expect(component.state().name).toBe('Not Mark');
+    // expect(component.state().name).toBe('Not Mark');
```
reginawang3495, line 123: Note: these are imperative tests, but it's I'm fine with that -- just letting you know
```suggestion
```

reginawang3495 (changes requested):
Looks mostly good to me--just a couple of nits!

src/components/common/ProfilePanel.js
```diff
@@ -26,58 +34,67 @@ import Switch from './Switch';
...
+  //States
+  const [nameIsHovering, setNameHover] = useState(false);
+  const [imageIsHovering, setImageHover] = useState(false);
+  const [editingName, setEditingName] = useState(false);
+  const [showModal, setShowModal] = useState(false);
+  const [nameSubmitted, setNameSubmitted] = useState(false);
+  const [name, setName] = useState(displayName);
+  const [selectedImage, setSelectedImage] = useState('');
+  const [displayNameMessage, setDisplayNameMessage] = useState('');
+  const [error, setError] = useState('');
```
reginawang3495, line 62: Nit: Add a comment saying that the error is never set

src/components/common/ProfilePanel.js
TrevorGuo: Yeah, name is the state while displayName is the prop. The name changing test failed when I used displayName rather than name. As far as I can tell, using either variable works as intended.

reginawang3495: Sounds good to me, this would be good to also test manually to make sure that unintended side effects happen in this change <since we also changed the tests itself>

(I already did but just noting for you)

reginawang3495 (approved):
looks good to me~ congrats on first teachla pull request!! 🎉 waow so cool

@krashanoff / @lumisphere902 Don't know if y'alls still active, but any thoughts? <since yalls are the code owners> We'll merge tomorrow if no response

krashanoff (approved):
I see no problems here at all. Thanks a ton for updating the unit tests, as well.

> Line 155 error was not declared previously as a state variable. I added it in, and set its initial value to an empty string.

I speculate this is because [`error`](https://github.com/uclaacm/TeachLAFrontend/pull/555/files#diff-dc4bc8b9577c6f349a4272ddd3e8c797ab221d88d2a47293a8ba6798b0cb8637R62) was originally planned to be initialized from props and passed down. Not something we should worry about now.

PR #610, opened 2021-09-27
remove border

lumisphere902:
Removes ugly border on SketchBoxe
Screen Shot 2021-09-27 at 10 26 14 AM
s


krashanoff (approved):
Looks good. Tangential but layout of our profile card buttons has also gone amiss.

buttons missing spacing

PR #611, opened 2021-09-28
Make Classes Page

lumisphere902:
Continuation of #399


src/components/Class/index.js
```diff
@@ -0,0 +1,355 @@
...
+    console.log('about to get class with this data: ' + JSON.stringify(data));
+    try {
+      fetch
+        .getClass(data, true, true)
+        .then((res) => {
+          if (!res.ok) throw new Error(`Error loading this class! Got status ${res.status}`);
+          return res.classData;
+        })
+        .then((json) => {
+          this.setState({
```
lumisphere902, line 113: Most of this doesn't belong in the state, will refactor later

lumisphere902 (commented):
Still WIP, but it somewhat functions

krashanoff:
Hey Timothy, you're a beast for finishing this PR out. It'll take me a day or two to review this, but I'll start tonight. Thanks much ^^

src/components/Error.js
```diff
@@ -23,7 +23,7 @@ class Error extends React.Component {
             Let Us Know!
           </Button>
           &nbsp;
-          <Link to="/">
+          <Link to={this.props.returnTo ? this.props.returnTo : "/"}>
```
krashanoff, line 26: Pretty sure we can just drop in:

```suggestion
          <Link to={this.props.returnTo || "/"}>
```

src/components/common/SketchBox.js
```diff
@@ -27,17 +41,17 @@ class SketchBox extends React.Component {
...
+            buttonData.filter(data => data.func)
+            .map(data => (
+              <Col className="p-2 text-center" onClick={data.func}>
+                <FontAwesomeIcon className="fa-lg" icon={data.icon} />
+              </Col>
+            ))
+            .reduce((acc, curr) => {
+              return acc.length > 0 ? [...acc, (<div className="sketch-button-divider" />), curr] : [curr];
+            }, [])
+          }
```
krashanoff, line 54: This is some really smart code, and it looks clean, but each child in a list should have its own `key`.

```suggestion
            .map((data, idx) => (
              <Col className="p-2 text-center" onClick={data.func} key={idx}>
                <FontAwesomeIcon className="fa-lg" icon={data.icon} />
              </Col>
            ))
            .reduce((acc, curr) => {
              return acc.length > 0 ? [...acc, (<div className="sketch-button-divider" />), curr] : [curr];
            }, [])
          }
```

krashanoff (changes requested):
Thanks a ton for all your work Timothy. The changes do function as expected, and it's a good start that's walled off by the `developerAcc` attribute, so I think it's okay to merge them after a few code quality changes. Past that, I'll just point out a few general thoughts for my review.

## Creating a class does not persist

Right now, if a developer account creates a class, it will not persist in their instructor's view of the "classes" page, but it *will* persist in the database, so students can still join it. Also, class data persists in associated student structs, which means that they can still be accessed in the frontend. Something is just going wrong in state management, I believe.

## Student's class page is missing information

When navigating to the student's view of a class page, there's some information missing like the instructor's name, student list, etc. Sketches are there, though.

class page missing instructor name and student list

## Comments are lacking

This code could benefit from comments. The basic UI construction flow is easy to follow, for example, but looking at the functions building it up or the redux actions inherited, for example, are a little trickier to keep pace with. For a dev inheriting these changes, it might be a rough learning curve.

***

Past these few points, this is looking like a solid start that can be elaborated on post-merge. I left a couple nitpicky notes in the thread, from when I started off reviewing by going line-by-line.

krashanoff:
Definitely take this time to clean up the code as best you can so that others can contribute. I reviewed as much as I could before the rebase, but I can't easily track the changes now, so I leave it in your hands.

Issue #623, opened 2021-10-06
Refactor src/components/Main.js to a functional component w/ hooks


lumisphere902:
https://github.com/uclaacm/TeachLAFrontend/blob/master/src/components/Main.js

reginawang3495:
Closed by #656

PR #656, opened 2021-10-20
Refactor main to hooks

megang33:
#623


reginawang3495 (dismissed):
Almost there!! Thanks for contributing Megan :) Just some quick fixes

src/components/Main.js
```diff
@@ -18,130 +18,134 @@ import '../styles/Main.scss';
...
+      screenHeight={screenHeight}
       // view only trigger
       viewOnly={false}
       // pane
-      panelOpen={this.props.panelOpen}
-      pane1Style={this.state.pane1Style}
-      changePane1Style={(newStyle) => this.setState(newStyle)}
+      panelOpen={panelOpen}
+      pane1Style={pane1Style}
+      changePane1Style={(newStyle) => setPane1Style(newStyle)}
```
lumisphere902, line 101: Couple things here:
First, this can probably be simplified to just `setPane1Style` instead of a function which does the same thing
Second, this is a functionality change; you need to update the input to this function in `EditorAndOutput.js` as well

src/components/Main.js
```diff
@@ -18,130 +18,134 @@ import '../styles/Main.scss';
...
-          this.setState({ viewMode: CODE_ONLY });
-        }
+  useEffect(() => {
+    if (screenWidth <= EDITOR_WIDTH_BREAKPOINT) {
+      if (viewMode === CODE_AND_OUTPUT) {
+        setViewMode(CODE_ONLY);
       }
     }
-  }
+  }, [screenWidth]);
```
lumisphere902, line 56: This should probably include viewMode as well

lumisphere902 (changes requested):
LGTM other than a couple small errors

lumisphere902 (approved):
LGTM! Good work, and thanks for bearing through the lint issues.

PR #660, opened 2021-10-22
Enforce linting

timothycpoon:
Before, the pre-commit hook was not executing, and we were only running prettier instead of eslint/stylelint. This aims to fix that.

~~Also, the GitHub action lint-staged is not working properly... will look into that as well~~

Theoretically, this should require linting to be fixed for any changed files in a PR, which will improve code quality incrementally over time

Tested lint-staged/GitHub Action on `src/components/Class/containers/ClassPageContainer.js`


reginawang3495:
lgtm!! we should probably add prettier to husky on the `ts-starter` as well, nice catch

PR #684, opened 2021-10-30
improve lint action

lumisphere902:
Only lint files that were changed since branching from master, instead of directly comparing w/ master


reginawang3495:
I’m a bit unclear about this—what’s the use case / reasoning behind adding this? I don’t really see why if ppl add and remove in a different commit they should have to manually fix an untouched file

lumisphere902:
The reason behind this change was that when we were doing a direct diff, a PR that was out of date w/ `master` would have the changes from other branches linted. As the lint errors would have been fixed in the other commit, the current version would have lint errors, and prevent merging. This happened with #656, which is what prompted me to make the change in the first place

reginawang3495 (approved):
Sorry for the delay -- this totally slipped my mind. Sounds good to me though!! I guess we can teach people how to squash their commits if this need to happen then!

PR #686, opened 2021-11-01
refactor: changed Output into a functional component w/ hooks

### LINKED ISSUE #670: Refactor Output.js into a functional component w/ hooks
https://github.com/uclaacm/TeachLAFrontend/blob/master/src/components/Output/Output.js

lpizzinidev:
Closes #670

### DIFF
--- a/src/components/Output/Output.js
+++ b/src/components/Output/Output.js
@@ -1,156 +1,130 @@
 import { faPlay, faTerminal } from '@fortawesome/free-solid-svg-icons';
 
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import { Button } from 'reactstrap';
 import { OUTPUT_ONLY } from '../../constants';
-import OpenPanelButtonContainer from '../common/containers/OpenPanelButtonContainer.js';
-import ViewportAwareButton from '../common/ViewportAwareButton.js';
-import EditorRadio from '../TextEditor/components/EditorRadio.js';
+import OpenPanelButtonContainer from '../common/containers/OpenPanelButtonContainer';
+import ViewportAwareButton from '../common/ViewportAwareButton';
+import EditorRadio from '../TextEditor/components/EditorRadio';
 
 /** --------Props--------
  * None
  */
 
-class Output extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      // used for the refresh button
-      counter: 0,
-      run: 0,
-      showConsole: true,
+const Output = React.memo(
+  ({
+    viewMode,
+    viewOnly,
+    screenHeight,
+    vLanguage,
+    language,
+    code,
+    runResult,
+    updateViewMode,
+    isSmall,
+  }) => {
+    const [counter, setCounter] = useState(0);
+    const [run, setRun] = useState(0);
+    const [showConsole, setShowConsole] = useState(true);
+    const [firstLoad, setFirstLoad] = useState(true);
+
+    useEffect(() => {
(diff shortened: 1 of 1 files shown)


src/components/Output/Output.js
```diff
@@ -1,156 +1,130 @@
...
+    updateViewMode,
+    isSmall,
+  }) => {
+    const [counter, setCounter] = useState(0);
+    const [run, setRun] = useState(0);
+    const [showConsole, setShowConsole] = useState(true);
+    const [firstLoad, setFirstLoad] = useState(true);
+
+    useEffect(() => {
+      setFirstLoad(false);
```
lumisphere902, line 33: I don't think the `shouldComponentUpdate` function was refactored correctly. You should pass in a second argument to react memo. Additionally, the useEffect dependency array should correspond to line 43-45 in the old file.

src/components/Output/Output.js
```diff
@@ -1,156 +1,130 @@
...
-  }
-
-  //= =============React Lifecycle Functions===================//
-  shouldComponentUpdate = (nextProps, nextState) => {
-    if (this.state.showConsole !== nextState.showConsole) {
-      return true;
-    }
-
-    if (this.props.mostRecentProgram !== nextProps.mostRecentProgram) {
-      this.firstLoad = true;
```
lumisphere902, line 34: This functionality is not preserved

src/components/Output/Output.js
```diff
@@ -1,156 +1,130 @@
...
+
+    const renderRadio = () =>
+      viewMode === OUTPUT_ONLY && (
+        <div style={{ marginLeft: 'auto' }}>
+          <EditorRadio viewMode={viewMode} updateViewMode={updateViewMode} isSmall={isSmall} />
+        </div>
+      );
+
+    const toggleConsole = () => {
+      setShowConsole((prevShowConsole) => !prevShowConsole);
```
lumisphere902, line 84: this can just be `setShowConsole(!showConsole)`

src/components/Output/Output.js
```diff
@@ -1,156 +1,130 @@
...
-  };
-
-  render() {
+      >
+        <FontAwesomeIcon icon={faTerminal} />
+      </Button>
+    );
+
+    const runCode = () => {
+      setRun((prevRun) => prevRun + 1);
```
lumisphere902, line 100: this should just be the run state, instead of this function

lumisphere902 (changes requested):
Mostly good, there are just a few issues, mostly with the memoization

lpizzinidev:
@lumisphere902 
I've updated the component.
Let me know if there are other changes to be made.

One note: I didn't change my `runCode` function implementation because it handles the `run` state like before.
So I think that my refactoring is fine in that case.
Let me know if I'm missing something.

src/components/Output/Output.js
lumisphere902: `setFirstLoad` should be called when `mostRecentProgram` changes

src/components/Output/Output.js
```diff
@@ -1,156 +1,143 @@
...
-  constructor(props) {
-    super(props);
-    this.state = {
-      // used for the refresh button
-      counter: 0,
-      run: 0,
-      showConsole: true,
-    };
-    this.firstLoad = true;
+const compareProps = (prevProps, nextProps) => {
```
lumisphere902, line 15: `true` should be returned when the component should not update; you reversed it

lumisphere902 (changes requested):
Couple of elaborations; also there are a few lint changes needed

reginawang3495:
@lpizzinidev Would you like to resolve conflicts and we can merge in? Thanks!

PR #692, opened 2021-11-02
Login.js refactoring

ArushRam:
Removed Class, added Function component with useState in place of this.state and useEffect in place of componentDidMount and componentWillUnmount. Replaced all this references to useState references, changed all setState calls to direct calls of useState setter functions.

### DIFF
--- a/src/components/Login.js
+++ b/src/components/Login.js
@@ -49,7 +49,7 @@ function Login (props) {
    return () => {
      window.removeEventListener('resize', () => setDummy(!dummy));
    };
- });
+ }, []);
 
  const getSVG = () => (
    <svg
--- a/src/components/Login.js
+++ b/src/components/Login.js
@@ -5,7 +5,7 @@ import {
 } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import LoginImg1 from '../img/login1.svg';
 import LoginImg2 from '../img/login2.svg';
 import LoginImg3 from '../img/login3.svg';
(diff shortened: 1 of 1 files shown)


src/components/Login.js
```diff
@@ -40,123 +40,117 @@ const themeColors = {
   4: ['#FF94DB', 'white'],
 };
 
-class Login extends React.Component {
-  state = {
-    dummy: false,
-    index: Math.floor(Math.random() * 5),
-  };
+function Login {
```
doubleiis02, line 43: Since we are changing Login from a class to a function, we need to have a proper function declaration.
Try changing this line to `function Login(props) {`. We need the parentheses to indicate that it's a function, and we need a `props` parameter inside in order to access any props passed down to the function, such as `props.initialState` in line 114.

Side note: since the only thing that we are exporting in this file is the Login function, instead of writing `export default Login;` at the bottom, we can remove that line and instead export it at the top of the function declaration by stating `export default function Login(props) {` on this line. But this is not necessary!

doubleiis02 (changes requested):
Great job refactoring `Login` to a functional (hook-based) component! The reason why the checks aren't passing is because of a syntax error, which I commented below about. From now on, try to run `npm start` or `npm build` before pushing your changes to check locally if your code is running correctly. If there is any compilation error, running either of those two commands will show that error!

doubleiis02 (approved):
~~LGTM! Congrats on successfully completing your first PR~~🥳
~~Feel free to press squash and merge down below once the checks finish!~~

Oops, missed a small error; Timothy mentions it below

src/components/Login.js
```diff
@@ -40,123 +40,117 @@ const themeColors = {
...
-  // basically, when the window resizes, we should recalculate get SVG - the window parameters change!
-  componentDidMount = () => {
-    window.addEventListener('resize', () => this.setState({ dummy: !this.state.dummy }));
-  };
+ useEffect(() => {
+   window.addEventListener('resize', () => setDummy(!dummy));
+   return () => {
+     window.removeEventListener('resize', () => setDummy(!dummy));
+   };
+ });
```
lumisphere902, line 52: This should be useEffect(func, [])

lumisphere902 (changes requested):
LGTM except one small change

src/components/Login.js
doubleiis02: If you wanted to know why the array is needed:

- Giving it an empty array acts like componentDidMount as in, it only runs once. (**this is what we want**)
- Giving it no second argument acts as both componentDidMount and componentDidUpdate, as in it runs first on mount and then on every re-render.
- Giving it an array as second argument with any value inside, eg , [variable1] will only execute the code inside your useEffect hook ONCE on mount, as well as whenever that particular variable (variable1) changes.

lumisphere902 (approved):
LGTM after lint fixes

reginawang3495:
@ArushRam let me know if you needed help figuring out linting!

reginawang3495:
chatted w/ @ArushRam and decided to close since this is a duplicate of #701 and this hasn't been tested yet!

Issue #485, opened 2021-04-26
Refactor Login.js to a functional component w/ hooks


lumisphere902:
Refactor this file! https://github.com/uclaacm/TeachLAFrontend/blob/master/src/components/Login.js

PR #701, opened 2021-11-07
Refactor Login

peacebonez:
Resolves https://github.com/uclaacm/TeachLAFrontend/issues/485


src/components/Login.js
```diff
@@ -40,22 +40,18 @@ const themeColors = {
...
-    window.addEventListener('resize', () => this.setState({ dummy: !this.state.dummy }));
-  };
+  useEffect(() => {
+    window.addEventListener('resize', () => setDummy((_dummy) => !_dummy));
 
-  componentWillUnmount = () => {
-    window.removeEventListener('resize', () => this.setState({ dummy: !this.state.dummy }));
-  };
+    return () => window.removeEventListener('resize', () => setDummy((_dummy) => !_dummy));
+  }, [dummy]);
```
lumisphere902, line 52: Don't think this needs to be dependent on `dummy`

lumisphere902 (changes requested):
LGTM w/ just one small nit

src/components/Login.js
peacebonez: Copy that.  If not dependent on dummy, the dummy var is unused throughout the file.  Should I just set up lint-ignore to silence the warning that dummy is not used?

mattxwang: Just to jump in here - I think the original intent of the code was to force a component re-render. If you can create a custom hook and/or some sort of listener that triggers a re-render (and then gets the window size, etc.), you can remove the dependency on the `dummy` state variable.

mattxwang: Here's one common hook used for the same case: https://usehooks.com/useWindowSize/

Here's an interesting SO post on something similar; you may (?) have to use `useRef`: https://stackoverflow.com/questions/66588340/custom-hook-for-window-resize

src/components/Login.js
```diff
@@ -82,81 +78,80 @@ class Login extends React.Component {
...
+  return (
+    <div className="login-page-content">
+      <div className="login-page-content-container">
+        <div
+          className="bottom-right-toggle"
+          onClick={() => setIndex(Math.floor(Math.random() * 5))}
+          onKeyDown={(e) => {
+            if (e.key === 'Tab') return;
+            setIndex(Math.floor(Math.random() * 5));
+          }}
```
mattxwang, line 100: Might be worth making `setIndex(Math.floor(Math.random() * 5))` a one-liner `const` function!

mattxwang (commented):
Left some comments! I'm a bit far removed from the editor now, so feel free to merge without an approving review from me.

src/components/Login.js
peacebonez: Hey Matt, have a look at the solution I have in place to dynamically render the SVG, it's looking quite nice when adjusting the window height.

peacebonez: Ahh disregard, on larger screen heights the SVG appears buggy.  I'll have to take a closer look.

Screen Shot 2021-11-07 at 8 47 56 PM

peacebonez: Ok I fixed this bug.  If the SVG was being rendered and then the innerHeight was increased, the SVG would look manipulated because it was not growing to fit the innerHeight.  Now the svg is growing with innerHeight.

src/components/Login.js
```diff
@@ -79,84 +74,99 @@ class Login extends React.Component {
...
-              <span className="teachla-green">ACM Teach LA</span>
-            </a>
-            {' '}
-            with
-            {' '}
-            <FontAwesomeIcon className="beating-heart" icon={faHeart} />
+  const [svg, setSvg] = useState(getBackgroundSVG());
+  const updateSvg = () => {
+    setSvg(getBackgroundSVG());
+  };
```
mattxwang, line 90: I'm a bit confused as to why we need this state variable - don't all the state variables in `getBackgroundSVG` already rely on existing state, and thus automatically change on re-render? Right now, we're putting this JSX element into state, but it should be generated *from* state.

peacebonez: When the svg is in state, the background svg is moving in an interesting wavy way that I thought was the intended effect but it's possible that it's a bug.  I'll remove the state and push up shortly

lumisphere902 (approved):
LGTM. Thanks for working on this; lots of good code quality upgrades. @mattxwang thoughts?

src/components/Login.js
```diff
@@ -79,84 +75,90 @@ class Login extends React.Component {
...
         </linearGradient>
       </defs>
     </svg>
   );
+  // basically, when the window resizes, recalculates getBackgroundSVG - the window parameters change!
+  useEffect(() => {
+    window.addEventListener('resize', () => {
+      setWindowHeight(window.innerHeight);
+    });
+    return () => window.removeEventListener('resize', () => setWindowHeight(window.innerHeight));
```
mattxwang, line 92: Is this the proper way to remove the event listener? I'm realizing that I may have made a mistake when I wrote the code initially. The two different instances of `() => setWindowHeight(window.innerHeight)` are anonymous functions with different underlying references, right? 

If that's the case, this code just needs to look like:

```suggestion
    const resizeHandler = setWindowHeight(window.innerHeight);
    window.addEventListener('resize', resizeHandler);
    return () => window.removeEventListener('resize', resizeHandler);
```

mattxwang: Other than this, I'm good to merge. Not necessary to wait on me to approve as well, since Timothy's already given it the green light!

peacebonez: This does the trick in getting the SVG to update in the intended way but there is a buggy side effect that I've experienced before - If the page is rendered and the window height is increased, you get a buggy looking SVG like the one in the screencap:
Screen Shot 2021-11-11 at 7 48 07 PM

With this code tweak:
```
const resizeHandler = () => setWindowHeight(window.innerHeight);
    window.addEventListener('resize', resizeHandler);
    return () => window.removeEventListener('resize', resizeHandler); 
```
It removes the bug, but the SVG doesn't have the cool swooping effect.

lumisphere902 (approved):
Reiterating my approval. Thanks for working on this!

peacebonez:
Screen Shot 2021-11-13 at 1 21 35 PM
Hey guys, after rebasing I'm getting this linting error yelling at me about the function component.  Have you experienced this?  My local vs code is not indicating any linting errors.

mattxwang (dismissed):

src/components/Login.js
```diff
@@ -39,33 +39,29 @@ const themeColors = {
...
-  state = {
-    dummy: false,
-    index: Math.floor(Math.random() * 5),
-  };
-
-  // basically, when the window resizes, we should recalculate get SVG - the window parameters change!
-  componentDidMount = () => {
-    window.addEventListener('resize', () => this.setState({ dummy: !this.state.dummy }));
-  };
+const Login = ({ create, initialState }) => {
```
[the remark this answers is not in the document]
mattxwang: If you're curious, [here is the MDN doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) that describes the differences from a normal function expression and an arrow function.

mattxwang:
@peacebonez do you want to try running the ESLint autofixer locally with `npm run lint-js-fix`?

lumisphere902:
It should automatically run it for you when you try to commit... have you ran `npm install`?

mattxwang:
Oh also - @lumisphere902 why is the `package-lock.json` diff here so large? I would look at that before merging

lumisphere902:
Ah, that could also be the source of errors. We are using Node v16.x

peacebonez:
Hey guys sorry for the delay here (work + holidays), I'm going to run the suggested actions and get back to you.

peacebonez:
All checks passed!  I'm now running node 16.2.0, generated a clean package-lock.json file, and followed the linting conventions when creating a new React Hook.

mattxwang:
Just wanted to clarify - am I needed for review here? It's been a while since I've been heavily involved, but glad to lend a hand if needed.

lumisphere902:
You're good; just going to merge this one after checks pass

mattxwang:
Great work @peacebonez! This was a really hefty PR, and I think you've improved a lot of the bad code I wrote 3 years ago 😊

PR #716, opened 2021-11-10
Join class modal

JackTheVac:
First Test
Unable to test due to error "Unhandled Rejection (TypeError): Failed to fetch"


reginawang3495:
@lumisphere902 could you give me access/let me know how to join the classes flag? thanks!

src/components/Classes/components/JoinClassModal.js
```diff
@@ -1,89 +1,107 @@
...
-      error: "",
-      redirect: false,
-    };
-  }
-
-  closeModal = () => {
-    if (this.props.onClose && {}.toString.call(this.props.onClose) === "[object Function]") {
-      this.props.onClose();
+import React, { useState } from 'react';
+// import * as fetch from "../../../lib/fetch.js";
```
reginawang3495, line 2: hm this doesn't look right, probably shouldn't be commented out. I'd assume this is what's causing https://gyazo.com/15dcc9b74c72bea286e9330f57b136d0

src/components/Classes/components/JoinClassModal.js
```diff
@@ -1,89 +1,107 @@
...
+import ReactModal from 'react-modal';
+import { Redirect } from 'react-router-dom';
+
+import {
+  Button, Container, Row, Col, FormGroup, Label, Input,
+} from 'reactstrap';
+
+const JoinClassModal = (props) => {
+  const [wid, setWid] = useState('');
+  // const [_disableSubmit, setDisableSubmit] = useState("")
```
reginawang3495, line 12: hm we should be able to leave this in without errors because it starts with `_`

See https://eslint.org/docs/rules/no-unused-vars#argsignorepattern + https://github.com/uclaacm/TeachLAFrontend/blob/master/.eslintrc.js#L46

src/components/Classes/components/JoinClassModal.js
```diff
@@ -1,89 +1,107 @@
...
+              ? "We couldn't find that class. Please try again!"
+              : 'There was a problem joining the class, please try again!',
+          );
+          // this.setState({
+          //   disableSubmit: false,
+          //   error: (err === "404") ?
+          //     "We couldn't find that class. Please try again!" :
+          //     "There was a problem joining the class, please try again!",
+          // });
+          // console.log(err);
```
reginawang3495, line 101: `console.error(err)` should work? 

https://github.com/uclaacm/TeachLAFrontend/blob/master/.eslintrc.js#L122

reginawang3495 (changes requested):
Here's a quick pass at the code without testing it!! Looks pretty good so far :))

lmk if anything i mention doesn't work/you need help!

src/components/Classes/components/JoinClassModal.js
JackTheVac: with or without it still causes the problem in the ss D:
And when I try to run through ESlint with "import * as fetch from "../../../lib/fetch.js";", it throws an error

src/components/Classes/components/JoinClassModal.js
JackTheVac: Yup console.error works

src/components/Classes/components/JoinClassModal.js
lumisphere902: Is this locally? And how are you running the editor? It looks like the backend might not be up; try `npm run staging_start`

src/components/Classes/components/JoinClassModal.js
```diff
@@ -1,89 +1,107 @@
...
 
     return false;
   };
 
-  onSubmit = async (e) => {
+  const onSubmit = async (e) => {
     e.preventDefault();
 
-    if (this.badInput()) {
+    if (badInput()) {
```
lumisphere902, line 58: You can check for `disableSubmit` here; I must have missed it from Will's PR

src/components/Classes/components/JoinClassModal.js
lumisphere902: Also, note that the default value should be a bool, just for semantic reasons.

lumisphere902 (changes requested):
Overall, it looks great; just a couple of comments. Let me know if you have any questions!

src/components/Classes/components/JoinClassModal.js
JackTheVac: it still shows unused variable when I uncomment it, seems like the documentation is saying that you can ignore only if the variable is an argument, but this is a useState variable.

lumisphere902: I think you can just check for the flag on line 58, which should resolve the issue.

src/components/Classes/components/JoinClassModal.js
reginawang3495: Yea I tested it, and with it, it works for me

src/components/Classes/components/JoinClassModal.js
reginawang3495: ok nice catch, you can just add `varsIgnorePattern: '^_' ` to the `.eslintrc.js` if you'd like! lmk if you need help w this :))

Also to @lumisphere902, not really sure what 
> I think you can just check for the flag on line 58, which should resolve the issue. 
means?

src/components/Classes/components/JoinClassModal.js
JackTheVac: npm run staging_start works for me as well, so i uncommented it now

src/components/Classes/components/JoinClassModal.js
JackTheVac: @reginawang3495 yah can you show me where to put varsIgnorePattern: '^_' ? 
I'm kinda confused where to put it

src/components/Classes/components/JoinClassModal.js
lumisphere902, line 58:
```suggestion
    if (_disableSubmit || badInput()) {
```

src/components/Classes/components/JoinClassModal.js
lumisphere902: I was just thinking that if we actually used the flag like this: https://github.com/uclaacm/TeachLAFrontend/pull/716/files#r748608755, we could avoid the eslint issue. The eslintrc change would go here: https://github.com/uclaacm/TeachLAFrontend/blob/master/.eslintrc.js#L46

reginawang3495: I think it’d be good to put in the eslint anyhow!

JackTheVac: Yeah I added disableSubmit at the line and the eslint didn't check for it, so i think it should be fine

reginawang3495:
synced up w @JackTheVac -- will finish up on tuesday!!

JackTheVac:
This final version should work

src/components/Classes/components/JoinClassModal.js
```diff
@@ -1,89 +1,82 @@
...
-  constructor(props) {
-    super(props);
-    this.state = {
-      wid: '',
-      disableSubmit: false,
-      error: '',
-      redirect: false,
-    };
-  }
+const JoinClassModal = function ({ props }) {
```
lumisphere902, line 9:
```suggestion
function JoinClassModal({ props }) {
```

lumisphere902 (approved):
LGTM. Just one small change for a lint warning, but it's not a huge deal.

.eslintrc.js
```diff
@@ -43,7 +43,9 @@ module.exports = {
   rules: {
     'linebreak-style': ['error', 'unix'],
 
-    'no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
+    'no-unused-vars': ['error', { varsIgnorePattern: '^_', argsIgnorePattern: '^_' }],
+
+    // 'no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
```
reginawang3495, line 48: can we remove this maybe?

src/components/Classes/components/JoinClassModal.js
```diff
@@ -1,89 +1,82 @@
...
+  const [wid, setWid] = useState('');
+  const [_disableSubmit, setDisableSubmit] = useState(false);
+  const [error, setError] = useState('');
+  const [redirect, setRedirect] = useState(false);
 
-  closeModal = () => {
-    if (this.props.onClose && {}.toString.call(this.props.onClose) === '[object Function]') {
-      this.props.onClose();
+  const closeModal = () => {
+    const { onClose } = props;
```
reginawang3495, line 16: would you like to destructure all props outside of this method? so something like on line 14
```
const {onClose, uid, ...} = props
```

src/components/Classes/components/JoinClassModal.js
```diff
@@ -1,89 +1,82 @@
...
+    setError('');
+    setDisableSubmit(false);
   };
 
-  badInput = () => {
-    if (!this.state.wid) {
-      this.setState({
-        error: 'Please enter the class code. You should get this from your instructor.',
-      });
+  const badInput = () => {
```
reginawang3495, line 27: this is just a cleaning up nearby code thing, but would you mind renaming this method to `isBadInput`? This makes it more clear that it returns a boolean!

reginawang3495 (approved):
looks good to me!!! thanks for refactoring~

just some small changes if you'd like to clean it up a bit :))

src/components/Classes/components/JoinClassModal.js
```diff
@@ -1,89 +1,82 @@
...
-    if (this.badInput()) {
+    if (badInput()) {
       return;
     }
 
     const data = {
-      uid: this.props.uid,
+      uid: props.uid,
       // Todo: swap this line back in when BE is fixed
       // wid: this.state.wid,
```
reginawang3495, line 47: @lumisphere902 is the backend fixed?

PR #736, opened 2021-11-15
converted CreateUserForm.js to functional components

huangdarren1106:


lumisphere902 (commented):
I’ll take a closer look later, but there are a lot of lint errors. If you merge in new changes from master, most should be auto fixed. Let me know if you run into any issues!

PR #737, opened 2021-11-15
refactor: fix as many lint errors as possible

reginawang3495:
## Summary
Ran `npm run lint-full-fix` so that new devs could just run this command and not have a ton of issues for linting!!

## Other
This will require force pushing, and might cause issues for current PRs but should be worth


lumisphere902 (approved):
LGTM, as this is just running the auto-fix. Can merge when tests and build pass

src/components/Classes/index.js
```diff
@@ -1,41 +1,39 @@
...
-      classCode: "",
+      selectedClass: '',
+      selectedImg: '',
+      selectedCid: '',
+      classCode: '',
     };
   }
 
-  componentDidMount = async () => {
+  componentDidMount() {
```
lumisphere902, line 36: Let's just revert this line
```suggestion
  componentDidMount = async () => {
```

PR #738, opened 2021-11-15
converted CreateUserForm.js to functional components

huangdarren1106:


reginawang3495:
@huangdarren1106 can you make a comment when you are ready for review? thanks!!

src/components/Login/CreateUserForm.js
```diff
@@ -41,33 +35,32 @@ export default class CreateUserForm extends React.Component {
    *    -Password characters: only alphanumeric characters, plus !@#$%
    * @return {boolean} - false when bad inputs given
    */
-  validateInputs = () => {
-    const { username, password, confirmPassword } = this.state;
+  const validateInputs = () => {
+    // const { username, password, confirmPassword } = state;
+    const username1 = username;
```
lumisphere902, line 40: What is the purpose of these 3 vars? I think you can just use `username`, etc

src/components/Login/CreateUserForm.js
```diff
@@ -166,71 +158,64 @@ export default class CreateUserForm extends React.Component {
     return addBreak ? <br /> : null;
   };
 
-  updateUsername = (username) => this.setState({ username });
+  const updateUsername = (newusername) => setusername(newusername);
```
lumisphere902, line 161: these 3 functions can be removed and replaced with the `set` functions

src/components/Login/CreateUserForm.js
```diff
@@ -1,36 +1,30 @@
...
-      errorMessage: '',
-      waiting: false,
-      usernameMessage: '',
-      passwordMessage: '',
-    };
-  }
+export default function CreateUserForm(props) {
+  // const init = props.initialState;
+  const { initialState } = props;
+  const [username, setusername] = useState(initialState ? initialState.username : '');
```
lumisphere902, line 21: For consistency's sake, let's name these in camelCase, e.g. `setUsername`

src/components/Login/CreateUserForm.js
```diff
@@ -166,71 +158,64 @@ export default class CreateUserForm extends React.Component {
...
-      borderColor: this.props.themeColor,
+      backgroundColor: themeColor,
+      borderColor: themeColor,
       borderWidth: 'medium',
       borderRadius: '4px',
-      color: this.props.textColor,
+      color: textColor,
     };
+
+    const [hoverButton, sethoverButton] = useState();
```
lumisphere902, line 210: This should be in the main component body, for consistency and so that all the state variables are easily visible.

lumisphere902 (changes requested):
Good work on the refactor! Based on the test failure, I think you might have missed a `setState` call(s). Left a couple of other comments as well; let me know if you have any questions :)

PR #740, opened 2021-11-17
Added tests and snapshots for DropdownButton

### LINKED ISSUE #619: Add tests for src/components/common/DropdownButton.js
As a common component, the DropdownButton should be tested. This should include what is rendered, the callback, etc.

https://github.com/uclaacm/TeachLAFrontend/blob/master/src/components/common/DropdownButton.js

jaczhi:
Resolves #619 

Added basic tests for DropdownButton that mainly use the snapshot feature to ensure all dropdown items are there and their elements are presented properly.


src/components/common/DropdownButton.test.js
```diff
@@ -0,0 +1,129 @@
...
+      />,
+    );
+    expect(componentFalse).toMatchSnapshot();
+  });
+
+  it('handles dropdown click correctly', () => {
+    const componentOpen = shallow(
+      <DropdownButton icon="lungs" DropdownItems={validDropdownItems} displayValue="foo" />,
+    );
+    componentOpen.find(DropdownToggle).simulate('click');
```
lumisphere902, line 73: Small comment: maybe we could check that it also closes? (If you click it again)

lumisphere902 (commented):
Great work on these test cases! I just have one small idea that you can choose to add or not. Otherwise, I think these tests work just fine

src/components/common/DropdownButton.test.js
jaczhi: Yeah for sure, that makes sense 

I added it to the test file

src/components/common/DropdownButton.test.js
```diff
@@ -0,0 +1,133 @@
...
+        icon="lungs"
+        DropdownItems={validDropdownItems}
+        defaultOpen={false}
+        displayValue="foo"
+      />,
+    );
+    expect(componentFalse).toMatchSnapshot();
+  });
+
+  it('handles dropdown click correctly', () => {
```
lumisphere902, line 69: OK, I don't think that the clicks are going through; I believe the clicks are handled in a way where shallow rendering would not allow that to happen. This could be fixed by mounting the component; however, I don't think this is necessary, as this is purely functionality implemented by the library anyway. I think this test can just be removed.

lumisphere902 (approved):
LGTM; there is a test that I think should be removed, as it is currently not functional. However, as it was basically intended to just test the library's functionality, I think it can just be removed.

PR #748, opened 2021-11-18
Allow lint errors to be committed; remove eslint rule

lumisphere902:
- Still runs lint on commit, but doesn't prevent commits on errors
- Remove the functional component rule

Note:
Screen Shot 2021-11-17 at 9 59 52 PM


reginawang3495 (approved):
LGTM thanks tim :)

PR #751, opened 2021-11-19
Refactored LoginForm.js

ArushRam:
Refactored LoginForm.js by changing LoginModal to functional component, fixed lint errors due to Login.js.

### DIFF
--- a/src/components/Login/LoginForm.js
+++ b/src/components/Login/LoginForm.js
@@ -1,6 +1,6 @@
 import SHA256 from 'crypto-js/sha256';
 import firebase from 'firebase/app';
-import React from 'react';
+import React, { useState } from 'react';
 import { Link } from 'react-router-dom';
 import { RingLoader } from 'react-spinners';
 import { Button } from 'reactstrap';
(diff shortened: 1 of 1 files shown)


src/components/Login/LoginForm.js
```diff
@@ -68,18 +64,20 @@ export default class LoginModal extends React.Component {
...
         });
     } else {
-      this.setState({ waiting: false, errorMsg: 'Failed to reach Firebase login services' });
+      setWaiting(false);
+      setErrorMsg('Failed to reach Firebase login services');
     }
   };
 
-  updateUsername = (username) => this.setState({ username });
+  const updateUsername = (myUsername) => setUsername(myUsername);
```
lumisphere902, line 76: These 2 functions can be removed, as they do the exact same thing as the set functions. For example, `updateUsername` can be removed and all its calls can be switched to `setUsername`

src/components/Login/LoginForm.js
```diff
@@ -9,25 +9,21 @@ import LoginInput from './LoginInput';
 import 'firebase/auth';
 import '../../styles/Login.scss';
 
-export default class LoginModal extends React.Component {
-  constructor(props) {
-    super(props);
+export default function LoginModal(props) {
+  const { themeColor, textColor } = props;
```
lumisphere902, line 13: Since these props are only used in one function, you can move the destructuring statement to within those functions. Since this component doesn't have a lot of props, this is optional, but can be helpful for clarity in larger components

lumisphere902 (changes requested):
Hey Arush, thanks for working on this! Most of it is good; there are just a couple of comments that you might want to address. Let me know if you have any questions!

lumisphere902 (approved):
LGTM

ArushRam:
> LGTM

Thanks for the feedback earlier!

PR #759, opened 2021-11-22
Refactored Output.js component to avoid using an <iframe> when running Python code

8BitRobot:
Having a bit of trouble with a specific part of the output refactoring and I could really use some help on this.

**Issue:** The canvas intended to display the Turtle code rendering is hidden from view. As far as I can tell, the animations are being drawn because there's a significant delay between print statements that corresponds with the time taken to execute any specific line of Turtle code. However, the canvas itself is completely hidden from view despite not having any CSS rules that would make this the case.

**Cause:** This bug came about when trying to refactor the output component to avoid using `iframe`s. Of course, this meant that I'd have to write terminal output directly to a containing `div` (which works) and the `canvas` element's ID has to be hard-coded into Python.js. I don't believe the issue has to do with Skulpt not detecting the `canvas` element because when I put in an invalid `canvas` ID, Skulpt actually does raise an error; therefore, I can be sure when it doesn't raise the aforementioned error that it's successfully detecting the `canvas` and the issue lies elsewhere.

Any help would be greatly appreciated. Thanks!


src/components/Output/Output.js
```diff
@@ -137,20 +157,27 @@ class Output extends React.Component {
...
+          <div
+            id={`${this.state.counter} ${this.state.run}`}
+            key={`${this.state.counter} ${this.state.run}`}
+            className="editor-output-iframe"
+            style={{ height: `${this.props.screenHeight - 61}px` }}
+          >
+            <div id="output">
+              <textarea id="inner" readOnly></textarea>
+            </div>
+            <canvas id="my-canvas"></canvas>
```
lumisphere902, line 173: Hey @8BitRobot , you kind of got baited by the name, which is completely understandable. The canvas should be a `div`, not a `canvas`. Hopefully that should make everything easier for you :)

lumisphere902:
CC: @krashanoff @mattxwang

Just wanted to check if you guys see any issue with this approach before a lot more time is invested into this, as I was not involved with the initial implementation of this.

mattxwang:
> Just wanted to check if you guys see any issue with this approach before a lot more time is invested into this

Directly consuming Skulpt is definitely the correct approach. What else did you want me to comment on? I might not have the relevant context since I'm not in the TLA dev team anymore.

(there should be _some_ sort of more direct React/node binding other than still having to interact with the DOM. Not exactly sure where I saw this, but there are quite a few JS interpreters for Python that provide this functionality. 

> as I was not involved with the initial implementation of this.

To be honest, I wasn't as well - if we wanted the original opinion, we'd probably want to ping @connorbo97 (four years removed at this point).

src/util/languages/Python.js
```diff
@@ -1,3 +1,14 @@
+import Sk from 'skulpt';
+
+Sk.externalLibraries = {
+    turtle: {
+        path: "/src/util/languages/turtle.js"
+    }
+};
```
krashanoff, line 7: Hey there, this approach seems pretty intuitive to me. I tested it out and right now `Sk.TurtleGraphics` is properly included in the namespace, but nothing paints, which I assume is intended in the PR's current state.

I've not worked with Skulpt before, but I'm fairly certain that you can directly link its stdlib into the iframe or webpage, which includes `turtle`. This will probably avoid some weird issues with rolling our own, and also prevents bloating our codebase. See their [JSDoc](http://skulpt.org/docs/index.html) for more.

lumisphere902: I think this is actually unused right now, Turtle should be included by default

lumisphere902:
> Directly consuming Skulpt is definitely the correct approach. What else did you want me to comment on? I might not have the relevant context since I'm not in the TLA dev team anymore.

That's good; just wanted to make sure that the `iframes` and the suggested alternative `fil` didn't provide some necessary functionality for some reason.

lumisphere902:
> I finally got around to finishing this PR and making it ready for review.
This is still currently marked as a draft PR; is this ready for review or did you want to make some additional changes?

> The Processing.js, React, and HTML interpreters all still use an iframe; only the Python one has been transitioned over. This is because I can't find a reliable way to do it without iframes that doesn't make use of the `eval` function (which, if I remember correctly, has its own fair share of security issues), and unfortunately the editor I was looking at before for inspiration (https://github.com/fatiherikli/fil) has been deleted or made private. However, the editor linked above did not actually work for React or Processing. The only reliable online IDE for React and Processing that I know of is repl.it, and that's almost entirely cloud-based according to my [research](https://cloud.google.com/customers/repl-it).
Changing Python to use a non-iframe implementation is good enough "for now". It's great to see that you did a lot of research into alternative methods for the other languages, though! :)

> In other news, I've got some linter errors due to files that I've never touched. Am I allowed to fix those linting errors myself or is it bad practice to overstep the "boundaries" of my ticket?
If you didn't touch it no need to fix them. If you happen to want to fix them, make a separate PR. This is less because it's "overstepping", and mostly so that a bunch of unrelated changes don't turn into a monstrous PR.

lumisphere902 (changes requested):
Couple of functionality changes from the old version which I assume are unintentional:

- Resizing the "console" part of the output does not move the "result" part
- The code is automatically run on first load (not a huge deal, but it is a functionality change)

Let me know if these are very difficult to change or anything like that; I just noticed the issues

lumisphere902 (approved):
LGTM; thanks for implementing those changes! Looks like you fixed just one lint error in an unrelated file, and it's causing the system to ask you to fix all of them. I would just revert that one lint change, so the PR doesn't get too bloated. :)

reginawang3495 (approved):
noice :^0

can we also rename the title of the PR to something more relevant ? And also editing the description since the stuff there is no longer relevant

krashanoff:
Apologies that I haven't been involved in this review. Going off of what Timothy and Regina have said, I think this PR is more than ready for merge following nits. Thanks for sticking with it, Premkumar.

src/components/Output/Output.js
```diff
@@ -26,56 +27,70 @@ class Output extends React.Component {
...
+      viewOnly, vLanguage, language, code, runResult,
+    } = this.props;
+
+    const renderLanguage = viewOnly ? vLanguage : language;
+    const renderRunResult = viewOnly ? code : runResult;
+    const { showConsole } = this.state;
+
+    // if (this.firstLoad) {
+    //    return null;
+    // }
```
reginawang3495, line 71:
```suggestion
```

src/components/Output/Output.js
```diff
@@ -87,63 +102,114 @@ class Output extends React.Component {
...
+    return viewMode === OUTPUT_ONLY && <OpenPanelButtonContainer />;
+  };
+
+  renderIframe = (getSrcDoc) => {
+    const { counter, run } = this.state;
+    const { screenHeight } = this.props;
+
+    // check if getsrcdoc is a function
+    if (!getSrcDoc && {}.toString.call(getSrcDoc) === '[object Function]') {
+      // console.log('Null src doc function found');
```
reginawang3495, line 120: nit: either remove this line or do `console.info`

src/components/Output/Output.js
```diff
@@ -153,4 +219,7 @@ class Output extends React.Component {
   }
 }
 
+// {this.renderBanner()}
+// <div>{this.renderOutput()}</div>
+
```
reginawang3495, line 224: was this intended?
```suggestion
```

src/util/languages/Python.js
```diff
@@ -1,103 +1,52 @@
...
+    } else {
+      received = true;
     }
-
-    function builtinRead(x) {
-        if (Sk.builtinFiles === undefined || Sk.builtinFiles["files"][x] === undefined)
-                throw "File not found: '" + x + "'";
-        return Sk.builtinFiles["files"][x];
+    if (received) {
+      // mypre.value = `${mypre.value}> ${text}`;
```
reginawang3495, line 17: was this intended?
```suggestion
```

src/util/languages/Python.js
```diff
@@ -1,103 +1,52 @@
...
-      ${getPythonSrcDocSkulptScript(code)}
-      ${showConsole ? '<div id="output"> </div>' : '<div id="output" style="display:none;"> </div>'}
-      <div id="mycanvas"></div>
-    </body>
-  `;
+  }
+  // function closeConsole() {
+  //   const mypre = document.getElementById('inner');
+  //   mypre.style.display = 'none';
+  // }
```
reginawang3495, line 27: was this intended?

```suggestion
```

reginawang3495 (approved):
Looks great!! Thanks for the changes :))

Can you take a quick look at your PR and remove any unnecessary comments and such? Also can you update your description to be relevant the the PR summary / remove it since it condenses to the commit message? 
image

Feel free to merge once you are done with those!

PR #811, opened 2022-02-02
Fix props error

lumisphere902:
Fixing a small syntax error that slipped through


jonathanbcarlson (approved):
LGTM, remove those pesky curly braces!

PR #880, opened 2022-04-28
fixed default sketch missing bug

### LINKED ISSUE #659: After deleting a sketch that was the "mostRecentProgram", the default sketch is bugged
When this occurs, we should default to one of the existing sketches, or have some sort of error page if there are no sketches
Screen Shot 2021-10-22 at 12 50 18 PM

The root cause of this issue is that the local state is set incorrectly after creating a sketch. The reason this bug occurred was that the mostRecentProgram, as well as the state of the created sketch, was not set properly.

PierceCCH:
Fixes #659. Added piece of code to fix sketch bugging out when the sketch being deleted is mostRecentProgram and the first program in the list.


src/components/Sketches/components/ConfirmDeleteModal.js
```diff
@@ -48,7 +48,11 @@ const ConfirmDeleteModal = function (props) {
           // this next piece of code is a guard against deleting mostRecentProgram - if we do,
           // then we need to re-populate it with something different.
           if (sketchKey === mostRecentProgram && programKeys.size > 0) {
-            setMostRecentProgram(programKeys.get(0));
+            if (sketchKey === programKeys.get(0)) {
+              setMostRecentProgram(programKeys.get(1));
```
lumisphere902, line 52: Very small edge case, but if the size of `programKeys` is only 1, this will crash

lumisphere902 (changes requested):
Hey Pierce, sorry for the delay on the review. Overall it's a lot of strong work, and good catch on the additional bug! There is a small edge case that should be fixed, though.

Other than that, there are just the lint errors. Most should resolve themselves if you run: `npm run lint-js-changes`. I know they aren't related to your work, but we are trying to iteratively increase the code quality through linting by requesting that people fix lint issues in files they edit. Let me know if you have any questions!

src/components/Sketches/components/CreateSketchModal.js
```diff
@@ -132,27 +134,33 @@ let CreateSketchModal = function (props) {
...
+      console.error(err);
     }
     setDisableSubmit(true);
     setError('');
   };
 
   const renderSecondModal = () => {
     const icons = ThumbnailArray.map((val, index) => (
-      <figure className="sketches-gallery-item" key={val} onClick={() => setThumbnail(index)}>
+      <figure
```
lumisphere902, line 157:
```suggestion
      <figure // eslint-disable-line jsx-a11y/no-noninteractive-element-interactions
```

lumisphere902 (approved):
LGTM, thanks for all the hard work! As we discussed, the error should be somewhat OK to ignore. Feel free to merge when you can!

PR #952, opened 2022-10-18
Move fetch calls out of userDataReducer.js

### LINKED ISSUE #635: Move fetch calls out of userDataReducer.js
https://github.com/uclaacm/TeachLAFrontend/blob/master/src/reducers/userDataReducer.js

Reducers should never contain API calls, which is the case here. Among other problems, error handling in particular is nearly impossible with this setup. This ticket would involve moving the fetch requests to the sources of the action calls. An example is in this file: https://github.com/uclaacm/TeachLAFrontend/blob/master/src/components/common/ProfilePanel.js#L91.
The fetch call for `setDisplayName` should be made here instead of in the reducer. Similar changes should be made to usages of `setPhotoName` and `setMostRecentProgram`.

A perhaps cleaner alternative would be to add middleware along the lines of `redux-thunk` and move the fetch requests to the action creators. This would allow for less code repetition, but increase the number of dependencies.

MDZHX:
Fixes #635.

Fetch calls are moved out of userDataReducer. The rest of the logic (error handling, in particular) remains untouched.

The reason I kept the original logic is that I think it may not be a good idea for me to make too many assumptions about the desired behavior, before any additional discussion/confirmation. In addition, I also saw that there's error-handling logic that depends on the previous implementation in other places. 

That said, we can definitely further discuss any related refactorings in this PR, or we can open a new issue for this.

### DIFF
--- a/src/components/Sketches/containers/ConfirmDeleteModalContainer.js
+++ b/src/components/Sketches/containers/ConfirmDeleteModalContainer.js
@@ -1,6 +1,6 @@
 import { connect } from 'react-redux';
 import { deleteProgram } from '../../../actions/programsActions';
-import { setMostRecentProgram, setError } from '../../../actions/userDataActions.js';
+import { setMostRecentProgram } from '../../../actions/userDataActions.js';
 import ConfirmDeleteModal from '../components/ConfirmDeleteModal.js';
 
 const mapStateToProps = (state) => {
@@ -16,7 +16,6 @@ const mapStateToProps = (state) => {
 const mapDispatchToProps = (dispatch) => ({
   deleteProgram: (program, data) => dispatch(deleteProgram(program, data)),
   setMostRecentProgram: (value) => dispatch(setMostRecentProgram(value)),
-  setUserDataError: (errorMsg) => dispatch(setError(errorMsg)),
 });
 
 const ConfirmDeleteModalContainer = connect(
--- a/src/components/Sketches/containers/ConfirmDeleteModalContainer.js
+++ b/src/components/Sketches/containers/ConfirmDeleteModalContainer.js
@@ -1,6 +1,6 @@
 import { connect } from 'react-redux';
 import { deleteProgram } from '../../../actions/programsActions';
-import { setMostRecentProgram } from '../../../actions/userDataActions.js';
+import { setMostRecentProgram, setError } from '../../../actions/userDataActions.js';
 import ConfirmDeleteModal from '../components/ConfirmDeleteModal.js';
 
 const mapStateToProps = (state) => {
@@ -16,6 +16,7 @@ const mapStateToProps = (state) => {
 const mapDispatchToProps = (dispatch) => ({
   deleteProgram: (program, data) => dispatch(deleteProgram(program, data)),
   setMostRecentProgram: (value) => dispatch(setMostRecentProgram(value)),
+  setUserDataError: (errorMsg) => dispatch(setError(errorMsg)),
 });
 
 const ConfirmDeleteModalContainer = connect(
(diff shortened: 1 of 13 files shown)


src/components/Sketches/index.js
```diff
@@ -51,7 +49,10 @@ class Sketches extends React.Component {
   };
 
   setProgram = (name) => {
-    this.props.setMostRecentProgram(name);
+    const { uid } = this.props;
+
+    // eslint-disable-next-line react/destructuring-assignment
```
lumisphere902, line 54: I believe this can be removed

src/components/common/DropdownButton.js
```diff
@@ -17,6 +17,7 @@ import {
 
 const DropdownButton = function (props) {
```
lumisphere902, line 18:
```suggestion
function DropdownButton(props) {
```
nit for lint

src/components/common/containers/ProfilePanelContainer.js
```diff
@@ -17,8 +17,34 @@ const mapStateToProps = (state, ownProps) => ({
 const mapDispatchToProps = (dispatch, ownProps) => ({
   onThemeChange: ownProps.onThemeChange,
   collectUserPhoto: () => {},
-  setDisplayName: (name) => dispatch(setDisplayName(name)),
-  setPhotoName: (name) => dispatch(setPhotoName(name)),
+  setDisplayName: (name, uid) => {
+    try {
+      fetch
```
lumisphere902, line 22: I think a fetch import is missing here

src/components/Sketches/containers/ConfirmDeleteModalContainer.js
```diff
@@ -15,7 +15,18 @@ const mapStateToProps = (state) => {
 
 const mapDispatchToProps = (dispatch) => ({
   deleteProgram: (program, data) => dispatch(deleteProgram(program, data)),
-  setMostRecentProgram: (value) => dispatch(setMostRecentProgram(value)),
+  setMostRecentProgram: (value, uid) => {
+    try {
+      fetch
```
lumisphere902, line 20: I think a fetch import is missing here

src/components/Sketches/containers/CreateSketchModalContainer.js
```diff
@@ -1,15 +1,26 @@
...
 const mapStateToProps = (state) => ({
   uid: state.userData.uid,
 });
 
 const mapDispatchToProps = (dispatch) => ({
   addProgram: (program, data) => dispatch(addProgram(program, data)),
-  setMostRecentProgram: (value) => dispatch(setMostRecentProgram(value)),
+  setMostRecentProgram: (value, uid) => {
+    try {
+      fetch
```
lumisphere902, line 14: fetch import

lumisphere902 (changes requested):
Hey Zihan, this PR turned out to be pretty big, but you did a great job! There is only one functional issue; the fetch calls shouldn't be working because the fetch library is not currently being imported. I marked some of them, but there are more and I didn't want to make this review super bloated. Other than that, this should be a solid refactor PR!

src/components/Sketches/index.js
MDZHX: Hi Timothy, thank you for the review! I fixed the other issues just now, but I'm not very sure about what this one is referring to. Do you mean to remove the destructuring of `uid` and use `this.props.setMostRecentProgram(name, this,props.uid)` directly in the next line?

lumisphere902:
I believe the eslint disable is unnecessary.

lumisphere902 (approved):
Everything looks to be addressed, so feel free to merge at your earliest convenience. The refactor went pretty smoothly, and the overall code consistency also grew with the changes for linting purposes. Good job!

PR #961, opened 2022-11-15
Fix getSketch

### LINKED ISSUE #959: Fix getSketch
It seems that when we were making changes to the request structure, getSketch was neglected. It should be changed to return a promise, and its usages in ViewOnly.js should also be changed to reflect that change

MDZHX:
Fixes #959.

### DIFF
--- a/src/components/ViewOnly.js
+++ b/src/components/ViewOnly.js
@@ -48,35 +48,31 @@ class ViewOnly extends React.Component {
       await this.codeSaverHelper();
     }
 
-    try {
-      fetch
-        .getSketch(programid)
-        .then((res) => {
-          if (!res.ok) {
-            this.setState({ notfound: true });
-            return Promise.reject(res);
-          }
-          return res.json();
-        })
-        .then((sketch) => {
-          const lang = getLanguageData(sketch.language);
-          this.setState({
-            sketchName: sketch.name,
-            language: lang,
-            code: sketch.code,
-            thumbnail: sketch.thumbnail,
-            loaded: true,
-          });
-          setProgramCode(mostRecentProgram, sketch.code);
-          setProgramLanguage(mostRecentProgram, sketch.language);
-          runCode(sketch.code, lang);
-        })
-        .catch((err) => {
-          console.error(err);
+    fetch
+      .getSketch(programid)
+      .then((res) => {
+        if (!res.ok) {
+          this.setState({ notfound: true });
+          return Promise.reject(res);
+        }
+        return res.json();
+      })
+      .then((sketch) => {
+        const lang = getLanguageData(sketch.language);
+        this.setState({
+          sketchName: sketch.name,
+          language: lang,
+          code: sketch.code,
+          thumbnail: sketch.thumbnail,
+          loaded: true,
         });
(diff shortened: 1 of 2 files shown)


src/components/ViewOnly.js
```diff
@@ -1,10 +1,12 @@
+/* eslint-disable react/destructuring-assignment */
```
lumisphere902, line 1: We talked about it already, but generally you shouldn't disable eslint rules without a good reason.

src/components/ViewOnly.js
```diff
@@ -74,21 +86,27 @@ class ViewOnly extends React.Component {
...
-      this.setState({ notfound: true });
-      return;
+  codeSaverHelper = () => {
+    try {
+      fetch
+        .getSketch(this.props.mostRecentProgram)
+        .then((res) => {
+          if (!res.ok) {
+            this.setState({ notfound: true });
+            return;
```
lumisphere902, line 98: Same promise changes as above

lumisphere902 (changes requested):
Most of the refactor looks good, there are just a couple of errors with how promises are handled. Good job!

src/components/ViewOnly.js
```diff
@@ -32,115 +35,164 @@ class ViewOnly extends React.Component {
...
     if (this.savePrevProgram) {
       await this.codeSaverHelper();
     }
 
-    const { ok, sketch } = await fetch.getSketch(this.props.programid);
-
-    if (!ok) {
-      this.setState({ notfound: true });
-      return;
+    try {
```
lumisphere902, line 51: I'm not 100% sure, but I think all cases of bad things happening will be caught by the `.catch` function on the promise, so you don't need a separate try/catch system

src/components/ViewOnly.js
```diff
@@ -32,115 +35,164 @@ class ViewOnly extends React.Component {
...
-    );
+  }
 
-    if (!okOriginal) {
-      this.setState({ notfound: true });
-      return;
+  codeSaverHelper = () => {
+    const { mostRecentProgram } = this.props;
+
+    try {
```
lumisphere902, line 107: Same here

src/components/ViewOnly.js
```diff
@@ -32,115 +35,164 @@ class ViewOnly extends React.Component {
...
+          theme={theme}
           onThemeChange={this.onThemeChange}
         />
         <div className="editor" style={codeStyle}>
           <EditorAndOutput
             // view mode
-            viewMode={this.state.viewMode}
-            updateViewMode={(viewMode) => this.setState({ viewMode })}
+            viewMode={viewMode}
+            updateViewMode={(_viewMode) => this.setState({ viewMode: _viewMode })}
```
lumisphere902, line 176: nit: this could be named something else like `vm`, as the underscore usually implies the variable is unused

lumisphere902 (changes requested):
Hey Zihan, this looks way better; the promise handling is really clean now. There are still a couple of minor topics, but other than that this is almost good to merge!

lumisphere902 (approved):
LGTM! Good job on this Pr

PR #1034, opened 2023-04-19
Fix crash on non-ASCII Python output

### LINKED ISSUE #521: Handle unicode characters in Python sketches
This isn't a super pressing issue, but a teacher did encounter a crash when pasting code from google docs due to the smart quotes. The sketch should either work or give an appropriate error message, rather than crash the entire editor.

TomBinford:
Fixes #521 (Unicode support) and a circular import issue caught by lints. Python programs can now safely output Unicode.
image


mizlan (approved):
looks good!

mizlan:
merge at your convenience

PR #1046, opened 2023-05-26
Encode current program into URL as `/editor/:programid`, push to browser history

### LINKED ISSUE #746: feat: encode filename/program into `/editor` path, push to browser history
Currently, the entire code editor portion of the app is housed under `/editor`. While this works, it poses a few frustrations:

1. tightly couples our implementation to `mostRecentProgram`, which has caused many problems
2. breaks the browser's "back"/"forward" functionality: intuitively, I should expect the back button to take me "back" one sketch
3. makes it harder to bookmark certain programs, etc.

I suggest we encode the active sketch into the URL parameters (either as a query string, like `/editor?project=...`, or as a path parameter, like `/editor/...`). This is very similar to how view-only mode already works (the latter implementation), so hopefully this isn't too tricky!

Since this is a feature that changes behaviour, let me know what you think @lumisphere902. Also IMO this is a good second issue, etc. for an editor team member.

TomBinford:
Closes #746 
This makes the URL reflect the current program, which used to live only in `mostRecentProgram` in the Redux state. On page load, the URL acts as a source of truth to set `mostRecentProgram`. After that point, we want Redux to be the source of truth so the URL will change whenever `mostRecentProgram` is set. This conveniently leaves the existing calls to `setMostRecentProgram` untouched.

This is also integrated with browser history so switching between files will let the forward/back buttons work as you expect.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,6 +17,7 @@
         "cross-env": "^7.0.3",
         "crypto-js": "^4.1.1",
         "firebase": "^9.12.1",
+        "history": "^4.9.0",
         "immutable": "^4.0.0",
         "react": "^16.14.0",
         "react-codemirror2": "^7.2.1",
@@ -30,6 +31,7 @@
         "react-split-pane": "^0.1.92",
         "reactstrap": "^9.0.1",
         "redux": "^4.1.2",
+        "redux-thunk": "^2.4.2",
         "skulpt": "^1.2.0"
       },
       "devDependencies": {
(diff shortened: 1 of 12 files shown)


TomBinford:
Follow-up: if a user shares a link like `/editor/someprogramid`, we should redirect other users (anyone but the owner) who use that link to `/p/someprogramid`. I haven't tested what would happen currently but I know it doesn't redirect.

mizlan (approved):

PR #1094, opened 2023-10-31
Convert ViewportAwareButton JS files to TypeScript

LeroyBG:
Renamed ViewportAwareButton and ViewportAwareButtonTest to from .js to .tsx and converted function interface to Typescript


src/components/common/ViewportAwareButton.tsx
```diff
@@ -9,7 +9,15 @@ import { Button } from 'reactstrap';
 /** --------Optional props--------
  * any props you would apply to a reactstrap Button.
  */
-const ViewportAwareButton = function (props) {
+
+interface ViewportAwareButtonProps{
+  icon : React.JSX.Element,
+  text : string,
+  isSmall : boolean,
+  remainder : any,
```
mizlan, line 17: Could you remove the space before the colon on each of these 4 lines please? LGTM with that change.

mizlan (changes requested):

mizlan:
Oh also: didn't notice that you renamed the test file as well. Could you please revert that change for now (undo the renaming?) I'm not sure if you changed any of the contents of the file but let's not touch that for now, it might get finicky.

src/components/common/ViewportAwareButton.tsx
LeroyBG: Yup!

mizlan (approved):

PR #1102, opened 2023-11-11
convert radio.js to tsx

ChancellorO:


src/components/common/Radio.tsx
```diff
@@ -25,8 +25,32 @@ import '../../styles/Radio.scss';
  *                        changes handle click to be called with all selected values
  */
 
-export default class Radio extends React.Component {
-  constructor(props) {
+interface OptionItem {
+  display : string;
```
mizlan, line 29: Remove all spaces before colons on all lines.

mizlan: And before `?:` too.

src/components/common/Radio.tsx
```diff
@@ -25,8 +25,32 @@ import '../../styles/Radio.scss';
...
+  value : number;
+}
+interface OptionsArray extends Array<OptionItem>{}
+
+interface RadioProps {
+  defaultSelected ?: any;
+  allowMultipleSelected ?: boolean;
+  options : OptionsArray;
```
mizlan, line 37: can you use replace `OptionArray` with `OptionItem[]`?

mizlan: And correspondingly remove line 32?

mizlan (changes requested):
When I run your code locally, I get a problem where `this.props` is not defined.

PR #1117, opened 2024-01-31
[EXAMPLE PR] Began converting LoginInput to TS.

### LINKED ISSUE #1097: feat: Convert LoginInput to TypeScript

kesdlvi:
Closes #1097


mizlan:
Good! But I think we can do more specific than `any` for the last two props.

